PR #70, opened 2015-10-02
DO NOT MERGE (CTH-134) Message hops should include the server identity

ploubser:
In the past the server URI was hardcoded to be pcp:///server, the URI
used by clients when addressing the server.

Here we update the broker-uri fn to return the correct server URI based
on the cn in it's ssl-certificate.

This relies on a `get-server` function to be added to
trapperkeeper-jetty9's webrouting service. This allows us to get the
server name from the current context and look up the corresponding
ssl-certificate file and extract the cn from there.

As a result the default web-router.conf webserver.conf files changed to include
the server name.


ploubser:
Closing due to tracking from opening the repo. Will open a new one.

PR #71, opened 2015-10-05
(maint) change default port to 8142

richardc:
During development we were setting up our test brokers on port 8080 and 8090
(non-ssl and ssl).  Now the 'production' port will be, for a time, 8142.
Update the sample config to use 8142 and the tests to use 8143 (so we can run a
sample server and the tests concurrently)

### DIFF
--- a/test/puppetlabs/pcp/broker/service_test.clj
+++ b/test/puppetlabs/pcp/broker/service_test.clj
@@ -17,7 +17,10 @@
 (def broker-config
   "A broker with ssl and own spool"
   {:webserver {:ssl-host "127.0.0.1"
-               :ssl-port 8081
+               ;; usual port is 8142.  Here we use 8143 so if we're developing
+               ;; we can run a long-running instance and this one for the
+               ;; tests.
+               :ssl-port 8143
                :client-auth "want"
                :ssl-key "./test-resources/ssl/private_keys/broker.example.com.pem"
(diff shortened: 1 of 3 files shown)


test/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -17,7 +17,10 @@
 (def broker-config
   "A broker with ssl and own spool"
   {:webserver {:ssl-host "127.0.0.1"
-               :ssl-port 8081
+               ;; usual port is 8142.  Here we use 8143 so if we're developing
+               ;; we can run a long-running instance and this one for the
+               ;; tests.
+               :ssl-port 8143
```
kylog: 8143 is what the orchestrator listens on, so this could get confusing at some point.

richardc: So if people are going to be developing against the broker on a machine that's really running an orchestrator instance, then yes, they'll clash.  We need something we can use safely that won't clash with our `lein tk` broker, or any other port used for anything else ever.  I think we're doomed.

kylog: Although we are certainly doomed, is there a downside to using, say, 18142 here?

kylog: This is more controversial than I realized. Disregard.

PR #74, opened 2015-10-08
(CTH-134) Message hops should include the server identity

ploubser:
In the past the server URI was hardcoded to be pcp:///server, the URI
used by clients when addressing the server.

Here we update the broker-uri fn to return the correct server URI based
on the cn in it's ssl-certificate. The ssl-cert is determined by looking
up the server value in the :websocket route, or directly from the server
if there is only one configured.

### DIFF
--- a/src/puppetlabs/pcp/broker/service.clj
+++ b/src/puppetlabs/pcp/broker/service.clj
@@ -8,7 +8,7 @@
 
 (trapperkeeper/defservice broker-service
   [[:ConfigService get-in-config]
-   [:WebroutingService add-ring-handler add-websocket-handler]
+   [:WebroutingService add-ring-handler add-websocket-handler get-server]
    [:MetricsService get-metrics-registry]]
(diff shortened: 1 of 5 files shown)


src/puppetlabs/pcp/broker/service.clj
```diff
@@ -17,6 +17,9 @@
           delivery-consumers (get-in-config [:pcp-broker :delivery-consumers] 16)
           authorization      (get-in-config [:pcp-broker :authorization] {:accept {:default :allow}})
           inventory          (make-inventory)
+          ssl-cert           (if-let [cert (get-in-config [:webserver (keyword (get-server this :websocket)) :ssl-cert])]
```
richardc: This would be more obvious as

```
(if-let [server (get-server this :websocket)]
  (get-in-config [:webserver server :ssl-cert])
  (get-in-config [:webserver :ssl-cert]))
```

PR #99, opened 2016-01-12
(PCP-222) Correct the ordering of the default rule

richardc:
As this rule comes from a package it should have a sort order in
the range 400-600 so that users can override with rules earlier and augment
with rules later.


mruzicka:
LGTM

mruzicka:
:+1:

PR #100, opened 2016-01-19
(maint) explain more about the loglevels in use

richardc:
Here we add the logging level rationale from PCP-5 to the doc/logging.md where
it will be the most useful, and also call out which level each message will be
seen at.


parisiale:
Good improvement.
Perhaps we could add a link to docs/logging.md in docs/README.md.

richardc:
Good call, added at cc0975a

PR #101, opened 2016-01-20
PCP-195 Supply "in-reply-to" header when responding to a message

richardc:
The "in-reply-to" envelope property was initially added to the
inventory_response codepath in commit 92b66d1c.  Here we change the
behaviour of the broker to supply this property whenever we are
clearly responding to a message, allowing a client to correlate a
response with its original request.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -151,11 +151,12 @@
         (sl/maplog :trace {:type :message-expired-from-server}
                    "Server generated message expired.  Dropping")
         capsule)
-      (let [response_data {:id (:id message)}
-            response (-> (message/make-message)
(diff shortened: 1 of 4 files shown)


parisiale:
Looks good to me.
Are we going to do the same for destination_report in a separate PR? PCP-195 is including such message type as well. As for error messages, I guess we have a separate ticket to deal with them.

richardc:
I missed the destination_report change, good catch.

error_message is a little more difficult without restructuring the code, as we don't distinguish 'couldn't parse your message' from 'could parse it, failed to handle it'.   In the former case we wouldn't have an id, in the latter we might.

richardc:
@parisiale I've added the missing case (destination report) and reworked the on-message flow so we can supply in-reply-to in 2 of the 3 error cases and left future todos to more clearly distinguish those error conditions.

parisiale:
:+1:

PR #103, opened 2016-02-03
(maint) Remove naming conflicts with PuppetDB

nicklewis:
This commit renames the namespaces copied from PuppetDB so that they are
unique to this project. This avoids conflicts with PuppetDB that arise
from having two (now different!) namespaces with the same name.

Similarly, changed the name of the embedded broker from "localhost" to
"pcp" to allow for multiple ActiveMQ instances to be started in the same
process.

These changes are meant to enable pcp-broker to be embedded in the same
jar/process as PuppetDB. Eventually this code probably ought to be
migrated to shared libraries or a dedicated trapperkeeper service (for
ActiveMQ), but for now this avoids conflicts.

### DIFF
--- a/src/puppetlabs/pcp/broker/activemq.clj
+++ b/src/puppetlabs/pcp/broker/activemq.clj
@@ -2,7 +2,7 @@
   (:require [clamq.protocol.connection :as mq-conn]
             [clamq.protocol.consumer :as mq-cons]
             [puppetlabs.pcp.broker.capsule :as capsule :refer [CapsuleLog]]
-            [puppetlabs.puppetdb.mq :as mq]
+            [puppetlabs.pcp.broker.borrowed.mq :as mq]
             [puppetlabs.structured-logging.core :as sl]
             [schema.core :as s]
             [taoensso.nippy :as nippy])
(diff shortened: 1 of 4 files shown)


richardc:
We do have plans to sync with puppetdb and turn this into a common trapperkeeper service, so I'm OK with this as a workaround in the interim.  However I'd rather we make the resulting namespaces continue to look odd and wrong, so it sticks out as 'this is stuff that needs to be cleaned up`, as `puppetlabs.pcp.broker.mq` looks like it's meant to be there.

To that end I'd propose instead of `puppetlabs.puppetdb.mq` -> `puppetlabs.pcp.broker.mq` we make it something like `puppetlabs.puppetdb.mq` -> `puppetlabs.pcp.broker.borrowed.puppetdb.mq` and `puppetlabs.puppetdb.cheshire` -> `puppetlabs.pcp.broker.borrowed.puppetdb.cheshire`.

nicklewis:
Updated to use `puppetlabs.pcp.broker.borrowed.cheshire` and `puppetlabs.pcp.broker.borrowed.mq` with namespace docs specifying that they were borrowed from PuppetDB.

richardc:
Cheers for this.

PR #104, opened 2016-02-04
PCP-250 Add multiple protocol version support

richardc:
We dub the developing version of the protocol "vNext" for now, as
it is a developing set of changes to the already established PCP
v1.0, and give it the websockets route :vNext.

Due to the closeness of v1.0 and vNext (vNext currently just adds
the optional in-reply-to header to all reply types), the brokers
internal representation of messages is vNext, which we then transform
when sending/receiving message via a Codec which is a property of
the connection derived from the uri the connection was established
on.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -64,7 +64,7 @@ webserver: {
 }
 ```
 
-The broker and the status services will need to be mounted using a
+The brokers protocol handlers and the status service will need to be mounted using a
 [webrouting](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-config.md)
 configuration.
 
@@ -73,6 +73,7 @@ web-router-service: {
     "puppetlabs.trapperkeeper.services.status.status-service/status-service": "/status"
     "puppetlabs.pcp.broker.service/broker-service": {
        v1: "/pcp"
+       vNext: "/pcp/vNext"
     }
 }
 ```
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -6,7 +6,7 @@
             [puppetlabs.experimental.websockets.client :as websockets-client]
             [puppetlabs.pcp.broker.activemq :as activemq]
             [puppetlabs.pcp.broker.capsule :as capsule]
-            [puppetlabs.pcp.broker.connection :as connection :refer [Websocket ConnectionState]]
+            [puppetlabs.pcp.broker.connection :as connection :refer [Websocket Codec ConnectionState]]
             [puppetlabs.pcp.broker.metrics :as metrics]
             [puppetlabs.pcp.message :as message :refer [Message]]
             [puppetlabs.pcp.protocol :as p]
@@ -74,8 +74,8 @@
 ;; connection map lifecycle
 (s/defn ^:always-validate add-connection! :- Connection
   "Add a Connection to the connections to track a websocket"
-  [broker :- Broker ws :- Websocket]
-  (let [connection (connection/make-connection ws)]
+  [broker :- Broker ws :- Websocket codec :- Codec]
+  (let [connection (connection/make-connection ws codec)]
     (swap! (:connections broker) assoc ws connection)
     connection))
 
@@ -218,7 +218,8 @@
   [broker :- Broker capsule :- Capsule]
   (if-let [websocket (get-websocket broker (:target capsule))]
     (try
-      (let [connection (get-connection broker websocket)]
+      (let [connection (get-connection broker websocket)
+            encode (get-in connection [:codec :encode])]
         (sl/maplog :debug (merge (capsule/summarize capsule)
                                  (connection/summarize connection)
                                  {:type :message-delivery})
@@ -226,7 +227,7 @@
         (locking websocket
           (time! (:on-send (:metrics broker))
                  (let [capsule (capsule/add-hop capsule (broker-uri broker) "deliver")]
-                   (websockets-client/send! websocket (capsule/encode capsule))))))
+                   (websockets-client/send! websocket (encode (capsule/encode capsule)))))))
       (catch Exception e
         (sl/maplog :error e
                    {:type :message-delivery-error}
@@ -287,6 +288,7 @@
         id       (:id request)
         uri      (:sender request)
         ws       (:websocket connection)
+        encode (get-in connection [:codec :encode])
         reason   (reason-to-deny-association broker connection uri)
         response (if reason {:id id :success false :reason reason} {:id id :success true})]
     (s/validate p/AssociateResponse response)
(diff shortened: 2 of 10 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -218,7 +218,9 @@
   [broker :- Broker capsule :- Capsule]
   (if-let [websocket (get-websocket broker (:target capsule))]
     (try
-      (let [connection (get-connection broker websocket)]
+      (let [connection (get-connection broker websocket)
+            {:keys [codec]} connection
+            {:keys [encode]} codec]
```
mruzicka: might be more readable to write:
```clojure
codec (:codec connection)
encode (:encode codec)
```

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -287,6 +289,8 @@
         id       (:id request)
         uri      (:sender request)
         ws       (:websocket connection)
+        {:keys [codec]} connection
+        {:keys [encode]} codec
```
mruzicka: same as above:
```clojure
codec (:codec connection)
encode (:encode codec)
```

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -419,38 +423,42 @@
...
-                              "Message {messageid} for {destination} from {commonname} {remoteaddress}")
-                   (let [next (determine-next-state broker capsule connection)]
-                     (swap! (:connections broker) assoc ws next)))))
+         (let [connection (get-connection broker ws)
+               {:keys [codec]} connection
+               {:keys [decode]} codec]
+           (try+
+             (let [message (decode bytes)]
+               (try+
+                 (if (not (check-sender-matches message connection))
```
mruzicka: ```(if-not )``` ?

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -419,38 +423,42 @@
...
+                   (send-error-message message "Message not authorized" connection)
+                   (let [uri (broker-uri broker)
+                         capsule (capsule/wrap message)
+                         capsule (capsule/add-hop capsule uri "accepted")]
+                     (sl/maplog :trace (merge (connection/summarize connection)
+                                              (capsule/summarize capsule)
+                                              {:type :connection-message})
+                                "Message {messageid} for {destination} from {commonname} {remoteaddress}")
+                     (let [next (determine-next-state broker capsule connection)]
+                       (swap! (:connections broker) assoc ws next))))
```
mruzicka: Just a note: I'd probably use the threading macro here:
```clojure
(->> (determine-next-state broker capsule connection)
     (swap! (:connections broker) assoc ws))
```

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -502,6 +510,21 @@
...
+(s/def default-codec :- Codec
+  {:decode message/decode
+   :encode message/encode})
+
+(s/def v1-codec :- Codec
+  "Codec for handling v1.0 messages"
+  {:decode message/decode
+   :encode (fn [message]
+             ;; strip in-reply-to for everything but inventory_response
+             (let [{:keys [in-reply-to message_type]} message
```
mruzicka: ```in-reply-to``` seems to be unused.

mruzicka: once removed the destructuring could be rewritten as:
```clojure
(let [message_type (:message_type message)]
  ...)
```

doc/configuration.md
```diff
@@ -64,7 +64,7 @@ webserver: {
 }
 ```
 
-The broker and the status services will need to be mounted using a
+The brokers protocol handlers and and the status service will need to be mounted using a
```
mruzicka: double ```and```

mruzicka:
Aside from the unused ```in-reply-to``` binding and the duplicate ```and``` in  comment this LGTM.
All the other comments are just about style and have no functional impact.

doc/configuration.md
richardc: Oops

src/puppetlabs/pcp/broker/core.clj
richardc: Maybe as 

```
encode (get-in connection [:codec :encode])
```

As we don't actually use codec.

src/puppetlabs/pcp/broker/core.clj
richardc: OK, but not this PR.

src/puppetlabs/pcp/broker/core.clj
richardc: Looks better, will propose a PR for that separately as all that's really changing here is the indent level.

src/puppetlabs/pcp/broker/core.clj
richardc: Good catch.  Bad leftover from an earlier spike implementation where I dissoced then conditionally assoced.

richardc:
Review comments addressed.  I have a branch for the style changes to on-message https://github.com/richardc/pcp-broker/tree/PCP-250-multiversion_style which I'd prefer to propose independently.

test/utils/puppetlabs/pcp/testutils.clj
```diff
@@ -0,0 +1,10 @@
+(ns puppetlabs.pcp.testutils
+  (:require [clojure.test :refer :all]))
+
+(defmacro dotestseq [bindings & body]
+  (if-not (seq bindings)
+    `(do ~@body)
+    (let [case-versions (remove keyword? (take-nth 2 bindings))]
+      `(doseq ~bindings
+         (testing (str "Testing case " '~case-versions)
```
mruzicka: I'm not sure if this is the intent, in any case the test context defined in this way results into the following context names:
```
Context Testing case (version)
```
and in some case to:
```
Context Testing case (sender-version receiver-version)
```

IMO we should include the actual versions - something like this:
```clojure
(testing (str "Testing case: " (clojure.string/join
                                 ", "
                                 (map #(clojure.string/join ": " %)
                                      (partition
                                        2
                                        (interleave '~case-versions (list ~@case-versions))))))
         ~@body)
```
which results into the following context names:
```
Context: Testing case: version: v1.0
Context: Testing case: version: vNext
Context: Testing case: sender-version: v1.0, receiver-version: v1.0
Context: Testing case: sender-version: v1.0, receiver-version: vNext
```

richardc: That's what I thought it already did.  I copied it from the puppetdb test suite so made the assumption that it was annotating with the values for the cases as it's way more useful that way.  I'll take this version.

mruzicka: :+1:

mruzicka:
+1 on submitting style changes separately
LGTM

PR #105, opened 2016-02-04
Style changes observed in the review of #104

richardc:
During the review of #104 these style enhancements were identified but not considered essential to the change at hand.  Here we clean those up.


mruzicka:
+1

PR #106, opened 2016-02-11
PCP-295 Track and return the service status

richardc:
Here we make the trapperkeeper-status endpoint more useful by actually tracking
the broker-service state so we can return a usable value to the server status
api.

    $ while true ; do  curl -sk https://127.0.0.1:8142/status/v1/services/broker-service | jq ".state" ; done
    "starting"
    "starting"
    "starting"
    "starting"
    "running"
    "running"

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -44,7 +44,8 @@
    :metrics-registry   Object
    :metrics            {s/Keyword Object}
    :transitions        {ConnectionState IFn}
-   :broker-cn          s/Str})
+   :broker-cn          s/Str
+   :state              Atom})
 
 (s/defn ^:always-validate build-and-register-metrics :- {s/Keyword Object}
   [broker :- Broker]
@@ -542,7 +543,8 @@
                               :uri-map            (atom {} :validator (partial s/validate UriMap))
                               :transitions        {:open connection-open
                                                    :associated connection-associated}
-                              :broker-cn          (get-broker-cn ssl-cert)}
+                              :broker-cn          (get-broker-cn ssl-cert)
+                              :state              (atom :starting)}
           metrics            (build-and-register-metrics broker)
           broker             (assoc broker :metrics metrics)]
       (add-websocket-handler (build-websocket-handlers broker v1-codec) {:route-id :v1})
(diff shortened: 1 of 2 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -542,7 +543,8 @@
                               :uri-map            (atom {} :validator (partial s/validate UriMap))
                               :transitions        {:open connection-open
                                                    :associated connection-associated}
-                              :broker-cn          (get-broker-cn ssl-cert)}
+                              :broker-cn          (get-broker-cn ssl-cert)
+                              :state (atom :starting)}
```
kylog: Whitespace nit but: align the value with the other values here?

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -551,22 +553,24 @@
...
     (mq/start-broker! activemq-broker)
-    (subscribe-to-queues! broker)))
+    (subscribe-to-queues! broker)
+    (reset! state :running)))
 
 (s/defn ^:always-validate stop
   [broker :- Broker]
-  (let [{:keys [activemq-broker activemq-consumers]} broker]
+  (let [{:keys [activemq-broker activemq-consumers state]} broker]
+    (reset! state :stopping)
```
kylog: Naive question but: is the server ever in an `:idle` state? Or is the thing after `:stopping` the termination of the service (i.e. such that there's no longer anything around to respond to the status endpoint)?

src/puppetlabs/pcp/broker/core.clj
richardc: Roger.

src/puppetlabs/pcp/broker/core.clj
richardc: There's no state after stopping. https://github.com/puppetlabs/trapperkeeper-status/blob/master/documentation/wire-formats.md  enumerates them as: "running", "error", "starting", "stopping", "unknown"

kylog: Ah perfect. I looked for the enumeration but somehow missed that, so cheers for the link.

kylog:
:+1:

mruzicka:
LGTM

PR #107, opened 2016-02-24
PCP-292 Add protocol-next feature switch

richardc:
Here we add a simple feature switch to enable the next protocol
mountpoint.  This will allow existing users to leave the developing
protocol support disabled until it becomes important enough to make
the configuration of it via webroutes mandatory.

As vNext isn't finalised/public yet, remove it from the default
web-routes.conf that ezbake will publish, and add a more specific
test-resources/conf.d/web-routes.conf.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -73,6 +73,7 @@ web-router-service: {
     "puppetlabs.trapperkeeper.services.status.status-service/status-service": "/status"
     "puppetlabs.pcp.broker.service/broker-service": {
        v1: "/pcp"
+       # vNext endpoint will need to be enabled with pcp-broker.protocol-vnext property
        vNext: "/pcp/vNext"
     }
 }
@@ -92,5 +93,8 @@ pcp-broker: {
 
     ## Number of consumers for the delivery queue.  Default is 16
     # delivery-consumers = 16
+
+    ## Enable the vNext protocol (needs corresponding webroute).  Default is false
+    # protocol-vnext = false
 }
 ```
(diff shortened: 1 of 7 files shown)


parisiale:
:+1:

doc/configuration.md
```diff
@@ -73,6 +73,7 @@ web-router-service: {
     "puppetlabs.trapperkeeper.services.status.status-service/status-service": "/status"
     "puppetlabs.pcp.broker.service/broker-service": {
        v1: "/pcp"
+       # vNext endpoint will need to be enabled with pcp-broker.protocol-vnext property
        vNext: "/pcp/vNext"
     }
```
mruzicka: I'm not too familiar with the ```web-router-service``` configuration structure, but this is different from what we have here: https://github.com/richardc/pcp-broker/blob/PCP-292_featureswitch_vnext/test-resources/conf.d/web-routes.conf#L7-L14

mruzicka:
Instead of introducing a new flag, could we base the decision whether to enable the next version protocol endpoint on the mere presence of the ```:vNext``` route in the ```web-router-service``` configuration?

doc/configuration.md
richardc: Yes, the one you're referring to in the link is using named webservers.

richardc:
The web-router-service configuration isn't ours, and if it changes how things are expressed in the future we'll be playing catchup.

richardc:
Actually I think there's an API we can use to detect this: https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-service.md#get-route

Will try it out that way on Monday and we can see how it stacks up.

richardc:
Going with the alternate implementation #108

PR #108, opened 2016-02-29
PCP-292 Only mount vNext when a web-route is supplied.

richardc:
Here we conditionally guard the mounting of the vNext protocol
handler behind it having a mountpoint present in the web-router
configuration.  This will allow existing users to leave the developing
protocol support disabled until it becomes important enough to make
the configuration of it via webroutes mandatory.

As vNext isn't finalised/public yet, remove it from the default
web-routes.conf that ezbake will publish, and add a more specific
test-resources/conf.d/web-routes.conf.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -68,6 +68,8 @@ The brokers protocol handlers and the status service will need to be mounted usi
 [webrouting](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-config.md)
 configuration.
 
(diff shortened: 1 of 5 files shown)


richardc:
@mruzicka  This is the alternate approach to gating the vNext protocol as suggested in the review comments of #107.

parisiale:
:+1:

mruzicka:
I prefer this one over https://github.com/puppetlabs/pcp-broker/pull/107

mruzicka:
LGTM

richardc:
Closing #107 then.

PR #109, opened 2016-02-29
PCP-301 update trapperkeeper to 1.3.0 for HUP

richardc:
Here we bump the trapperkeeper dependency up to 1.3.0 so in future
we can make use of the HUP handling added for TK-202.

In order to reduce some dependency clash, we no longer state our
dependencies on prismatic/schema, org.clojure/core.async, and
me.raynes/fs and instead resolve them transitively via
puppetlabs/trapperkeeper, and puppetlabs/kitchensink.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,5 @@
-(def tk-version "1.1.1")
-(def ks-version "1.1.0")
+(def tk-version "1.3.0")
+(def ks-version "1.3.0")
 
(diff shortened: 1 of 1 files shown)


parisiale:
Behavior already checked for PCP-247.
:+1:

mruzicka:
This is always a gamble but it resolves for me without warnings.

PR #110, opened 2016-03-04
PCP-294 Reject connections before we start

richardc:
If broker recieves a message or new connection while it is not in
the state :running, close the connection with code 1011:

https://tools.ietf.org/html/rfc6455#section-7.4

      1011 indicates that a server is terminating the connection because
      it encountered an unexpected condition that prevented it from
      fulfilling the request.

### DIFF
--- a/test/integration/puppetlabs/pcp/broker/service_test.clj
+++ b/test/integration/puppetlabs/pcp/broker/service_test.clj
@@ -15,7 +15,8 @@
             [puppetlabs.trapperkeeper.services.webserver.jetty9-service :refer [jetty9-service]]
             [puppetlabs.trapperkeeper.testutils.bootstrap :refer [with-app-with-config]]
             [puppetlabs.trapperkeeper.testutils.logging
-             :refer [with-test-logging with-test-logging-debug]]))
+             :refer [with-test-logging with-test-logging-debug]]
+            [slingshot.slingshot :refer [throw+ try+]]))
 
 (def broker-config
   "A broker with ssl and own spool"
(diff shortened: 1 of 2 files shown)


test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -92,6 +93,40 @@
...
+      (try+
+       (with-open [client (client/http-client-with-cert "client01.example.com")
+                   ws (http/websocket client
+                                      "wss://127.0.0.1:8143/pcp/vNext"
+                                      :close (fn [ws code reason] (swap! closed conj-unique code)))]
+         ;; Sleep so we can be closed by the broker, not us hitting the end of the with-open
+         (Thread/sleep 10))
+       (catch Object _
+         (swap! closed conj-unique :refused))))
+    (is (= [:refused 1011 1000 1011 :refused] @closed))
```
richardc: Travis runs are underlying this has a racy nature:

```
expected: (= [:refused 1011 1000 1011 :refused] (clojure.core/deref closed))
  actual: (not (= [:refused 1011 1000 1011 :refused] [:refused 1000 1011 1006 :refused]))
```

It's also possible that the outer-loop is too slow, so we get the sequence `[:refused 1011 1000 :refused]` or other permutations.

I think this test needs to use some form of state-machine, or some other way of expressing we may expect to transition forwards through the states.

richardc:
I refactored the test to allow for races to yielding multiple possible sequences, and so we now test for membership of the set of valid observable close-code sequences.  This test should be more robust when running in different environments.

parisiale:
:+1:

PR #112, opened 2016-03-18
PCP-245 check message expiry consistently

richardc:
These set of commits adds the 3 missing checks of message expiry; during association, once associated before persisting the message to the deliver queue, and an consumption of the message from the delivery queue.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -223,24 +223,26 @@
 (s/defn ^:always-validate deliver-message
   "Message consumer. Delivers a message to the websocket indicated by the :target field"
   [broker :- Broker capsule :- Capsule]
-  (if-let [websocket (get-websocket broker (:target capsule))]
-    (try
-      (let [connection (get-connection broker websocket)
-            encode (get-in connection [:codec :encode])]
-        (sl/maplog :debug (merge (capsule/summarize capsule)
-                                 (connection/summarize connection)
-                                 {:type :message-delivery})
-                   "Delivering {messageid} for {destination} to {commonname} at {remoteaddress}")
-        (locking websocket
-          (time! (:on-send (:metrics broker))
-                 (let [capsule (capsule/add-hop capsule (broker-uri broker) "deliver")]
-                   (websockets-client/send! websocket (encode (capsule/encode capsule)))))))
-      (catch Exception e
-        (sl/maplog :error e
-                   {:type :message-delivery-error}
-                   "Error in deliver-message")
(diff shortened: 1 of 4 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -267,16 +269,20 @@
 (s/defn ^:always-validate reason-to-deny-association :- (s/maybe s/Str)
   "Returns an error message describing why the session should not be
   allowed, if it should be denied"
-  [broker :- Broker connection :- Connection as :- p/Uri]
```
mruzicka: Just wondering what did "as" stand for?

richardc: Connected 'as' this uri I guess.  `uri` would have been better, or `sender`.

parisiale:
I don't think that treating a TTL expiration as an Association failure is semantically correct.
Clients won't be able to distinguish a true Association failure from a TTL expiration. And that is bad, as for the second case, they should retry the Session Association whereas for the first one they should terminate, as the problem would be due to a PCP misconfiguration.

On the other hand, I see the need to treat a TTL expiration during the Association in a different way. Could it be done by sending a PCP Error and then dropping the WebSocket?

In any case, this behaviour should be documented in the specs.

src/puppetlabs/pcp/broker/core.clj
mruzicka: ok, thanks

mruzicka:
LGTM

richardc:
I was following along with the spirit of [Association Message Flow](https://github.com/puppetlabs/pcp-specifications/blob/master/pcp/versions/1.0/association.md#association-message-flow)

> In case the request is processed successfully, the broker will reply with a associate response message.  The broker may ignore the request or reply with an /error message/ in case of node identity mismatch

But extending into reasons for the /error message/ that of recieving an expired association request.

We could instead say that an expired message must always result in a [ttl_expired](https://github.com/puppetlabs/pcp-specifications/blob/master/pcp/versions/1.0/ttl_expired.md) behaviour though this is also ambiguous as ttl_expired is documented in terms of message delivery.  Message delivery is a general capability of a broker, and [Session Association](https://github.com/puppetlabs/pcp-specifications/blob/master/pcp/versions/1.0/association.md) states:

> PCP clients must associate their session with a broker before they can use its capabilities

richardc:
@parisiale Have reworked this so we send ttl_expired and leave the connection open on expired association request.  The commit that changed is  0f10fedc -> 1e96a3e

parisiale:
:+1:

PR #114, opened 2016-04-18
(PCP-370) Remove "alternative" set of SSL certificates

parisiale:
Removing the "alternative" set of certificate as it's not used.


mruzicka:
LGTM

PR #116, opened 2016-04-27
(maint) Remove Support section from README

parisiale:
The "Support" section is redundant as we have now the "Maintenance"
section that explains how to fill tickets; with this commit, we remove
"Support".


mruzicka:
LGTM

PR #117, opened 2016-05-06
(PCP-162) add i18n support

mruzicka:
This commit adds i18n support to the project through the use of the [puppetlabs/i18n](https://github.com/puppetlabs/clj-i18n) library.


src/puppetlabs/pcp/broker/activemq.clj
```diff
@@ -20,7 +21,7 @@
     (sl/maplog :trace (assoc (capsule/summarize capsule)
                              :queue queue
                              :type :queue-enque)
-               "Delivering message {messageid} for {destination} to {queue} queue")
+               (i18n/trs "Delivering message '{messageid}' for '{destination}' to '{queue}' queue"))
```
parisiale: For clj-pxp-puppet's i18n we had "The single quotes need to be duplicated".
Isn't it the same here?

parisiale: Same for the other uses of single quote below.

src/puppetlabs/pcp/broker/borrowed/mq.clj
```diff
@@ -34,7 +35,7 @@
           (fn? usage-fn)
           (string? desc)]}
   (when megabytes
-    (log/info "Setting ActiveMQ " desc " limit to " megabytes " MB")
+    (log/info (i18n/trs "Setting ActiveMQ {0} limit to {1}MB" desc megabytes))
```
parisiale: Missing space before "MB".

src/puppetlabs/pcp/broker/borrowed/mq.clj
```diff
@@ -246,7 +246,7 @@
       (.readBytes bytes-message buffer)
       buffer)
     :else
-    (throw (ex-info (str "Expected a text message, instead found: " (class message))))))
+    (throw (ex-info (i18n/trs "Expected a text message, instead found: {0}" (class message)) {}))))
```
parisiale: Why the empty map?

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -128,7 +129,7 @@
                                :type :message-authorization
                                :allowed allowed
                                :message message)
-                 "Authorizing {messageid} for {destination} - {allowed}: {message}")
+                 (i18n/trs "Authorizing '{messageid}' for '{destination}' - '{allowed}': '{message}'"))
```
parisiale: Just to be in the safe side, does the structured login string interpolation work properly after the i18n translation?

src/puppetlabs/pcp/broker/activemq.clj
parisiale: We've just had a voice chat about this.

What we're doing here is making sure that the translation does not change what's inside the single quotes, so that structured login can interpolate afterwards.

@libbymo , do we really need to escape with single quotes here?

mruzicka: I introduced the single quotes to make sure the curly braces are not processed by [java.text.MessageFormat](https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html) as patterns. Perhaps this is not necessary, as the strings enclosed in the curly braces are not valid MessageFormat's patterns so they could be ignored even without quoting.

src/puppetlabs/pcp/broker/borrowed/mq.clj
mruzicka: The ```(ex-info )``` requires two arguments, when called with one it crashes:
```
(ex-info "now what")
ArityException Wrong number of args (1) passed to: core/ex-info  clojure.lang.AFn.throwArity (AFn.java:429)
```

src/puppetlabs/pcp/broker/core.clj
mruzicka: Will test that.

src/puppetlabs/pcp/broker/borrowed/mq.clj
mruzicka: It was intentional. Do we have some guidelines on this? - Whether to put space between a value and its unit or not?

src/puppetlabs/pcp/broker/borrowed/mq.clj
parisiale: It seems that it was broken, as `ex-info` signature is indeed `(ex-info msg map)    (ex-info msg map cause)`.
But, if there's no additional data (empty map), why are we throwing an ExceptionInfo?

src/puppetlabs/pcp/broker/borrowed/mq.clj
parisiale: Consistency.

src/puppetlabs/pcp/broker/borrowed/mq.clj
mruzicka: I have no idea. Perhaps it's caught somewhere, that's why I didn't want to change it.

src/puppetlabs/pcp/broker/activemq.clj
mruzicka: On the other hand if we wanted to use structured logging on the following map:
```clojure
(let [exception-map {:0 "java.lang.NullPointerException"
                     :1 "java.io.IOException"}]
  ...
  )
```
like so:
```clojure
(sl/maplog :info exception-map (i18n/trs "the ultimate cause was {0}"))
```
the quoting would be necessary to prevent ambiguity as to when the ```{0}`` is to be expanded.

mruzicka: It seems the quoting is indeed necessary:
```clojure
(puppetlabs.i18n.core/trs "now what? {text}")
NumberFormatException For input string: "text"  java.lang.NumberFormatException.forInputString (NumberFormatException.java:65)
```
vs.
```clojure
(puppetlabs.i18n.core/trs "now what? '{text}'")
=> "now what? {text}"
```

src/puppetlabs/pcp/broker/borrowed/mq.clj
parisiale: From what I just saw in our code, such exception is not caught anywhere. If I change the code to (correctly, with the empty map) throw it here, the error ends up in the JMS message listener. Here is what I get after connecting pxp-agent and sending a ping request from another client: https://gist.github.com/parisiale/fb98ef64138470b5e73485ece6fb23b7

Also, by looking where such function is called, it seems that the exception may end up propagated up to the `start()` function we pass to `trapperkeeper`, which should not throw: https://github.com/puppetlabs/trapperkeeper/blob/master/documentation/Error-Handling.md#errors-during-init-or-start

I may be missing something here as I'm not familiar with this code. c/c @MikaelSmith

src/puppetlabs/pcp/broker/activemq.clj
parisiale: 👍 
I suggest to document this in clj-i18n

libbymo: https://github.com/puppetlabs/clj-i18n#single-quotes-in-messages

A little late to the party! But yes, double quoting is necessary.

src/puppetlabs/pcp/broker/core.clj
mruzicka: Verified by providing a translation to https://github.com/puppetlabs/pcp-broker/pull/117/files#diff-1cf4fb148a15dbfdeda179f90790a86aR24 and starting the broker like this:
```shell
JVM_OPTS='-Duser.language=es -Duser.country=ES' lein tk
```

The message was correctly localized and interpolated:
```
2016-05-09 18:49:31,655 TRACE [DefaultMessageListenerContainer-1] [p.p.b.activemq] Entregando mensaje 3438cddd-3fb2-45f1-a867-9ddf17a6e654 para pcp://client01.example.com/puppet-api-test a la cola delivery
```

src/puppetlabs/pcp/broker/borrowed/mq.clj
```diff
@@ -246,7 +246,9 @@
       (.readBytes bytes-message buffer)
       buffer)
     :else
-    (throw (ex-info (str "Expected a text message, instead found: " (class message))))))
+    (throw (RuntimeException.
+             ^String (i18n/trs "Expected an object implementing either {0} or {1}, instead found object of class: {2}"
+                               (.getName javax.jms.TextMessage) (.getName javax.jms.BytesMessage) (-> message class .getName))))))
```
mruzicka: Since we don't catch this exception it doesn't make sense to use ```(ex-info )``` with no additional info map, a simple ```java.lang.RuntimeException``` will to the job. While at it, the massage is rephrased to more accurately describe why the Exception is thrown.

src/puppetlabs/pcp/broker/borrowed/mq.clj
```diff
@@ -246,7 +246,9 @@
       (.readBytes bytes-message buffer)
       buffer)
     :else
-    (throw (ex-info (str "Expected a text message, instead found: " (class message))))))
+    (throw (RuntimeException.
+             ^String (i18n/trs "Expected an object implementing either {0} or {1}, instead found: {2}"
+                               (.getName javax.jms.TextMessage) (.getName javax.jms.BytesMessage) (class message))))))
```
mruzicka: Since we don't catch this exception it doesn't make sense to use ```(ex-info )``` with no additional info map, a simple ```java.lang.RuntimeException``` will do the job. While at it, the massage is rephrased to more accurately describe why the Exception is thrown.

src/puppetlabs/pcp/broker/borrowed/mq.clj
MikaelSmith: I'm not particularly familiar with TK lifecycle. Reading the documentation it looks like a startup failure that can't be recovered from should cause the entire TK to shutdown, so this sounds like sane behavior.

I don't read TK's documentation as saying they shouldn't throw. It seems to say that if the service will fail to start, you should throw as early as possible so that TK shuts down quickly.

parisiale: Yes, but errors raised while handling messages with unexpected format should not led to a service stop. I need to spend more time on this.

parisiale:
Travis is not happy; 3 failures and 1 error.
I suggest to not touch the exceptions with this PR; if it's the case, we can make a separate ticket to ensure that errors are handled properly.

src/puppetlabs/pcp/broker/borrowed/mq.clj
MikaelSmith: Ah, then yes we probably want to catch rather than allowing it shutdown TK.

mruzicka:
All the errors/failures are in a single test - [send-to-never-connected-will-get-expired-test](https://github.com/mruzicka/pcp-broker/blob/PCP-162_localization/test/integration/puppetlabs/pcp/broker/service_test.clj#L369) - I wan't able to reproduce the failure on my machine. It's most probably not caused by the changed exception.

mruzicka:
I've re-triggered the build and it succeeded ... I'd blame it on a random error.

mruzicka:
I've put back the ```(ex-info )``` exception. (Which was briefly replaced with plain RuntimeException.)

src/puppetlabs/pcp/broker/activemq.clj
parisiale: Thanks @libbymo

PR #118, opened 2016-05-07
(maint) Bump trapperkeeper-webserver-jetty9 to 1.5.7

parisiale:
Bumping trapperkeeper-webserver-jetty9 to 1.5.7 as a recommended bugfix
is included; updating associated dependencies, including trapperkeeper
1.4.0. Note that, this time, we're not bumping clojure to 1.7.0.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def tk-version "1.3.0")
+(def tk-version "1.4.0")
 (def ks-version "1.3.0")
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Why are you sticking with Clojure 1.6.0?

parisiale:
To avoid making other updates with this PR, that tackles only the memory leak issue. Also, it would be better to update clojure in all components altogether. I'll make a ticket to track this.

MikaelSmith:
👍

PR #120, opened 2016-05-31
(PCP-448) Use ConcurrentHashMap for tracking connections

MikaelSmith:
Many Jetty threads can result in modifying the connections or uri map.
When many updates occur concurrently, we can see lots of failed atom
updates resulting in much slower processing.

Switch to using a ConcurrentHashMap. It appears to scale better, and is
at least as fast for smaller numbers of clients.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -22,15 +22,8 @@
             [puppetlabs.i18n.core :as i18n])
   (:import (puppetlabs.pcp.broker.capsule Capsule)
            (puppetlabs.pcp.broker.connection Connection)
-           (clojure.lang IFn Atom)))
-
-(def UriMap
-  "Mapping of Uri to Websocket, for sending"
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Tested this using a test setup of
```
Connection test setup:
  4 concurrent sets (+0 per run) of 100 endpoints (+50 per run)
  3 runs, 60000 ms pause between each run
  200 ms pause between each endpoint connection
  WebSocket connection timeout 2500 ms; Association Request TTL 5 s
```
where it took nominally the same time before and after (14m21s vs 14m9s), and
```
Connection test setup:
  4 concurrent sets (+0 per run) of 400 endpoints (+100 per run)
  3 runs, 60000 ms pause between each run
  200 ms pause between each endpoint connection
  WebSocket connection timeout 2500 ms; Association Request TTL 5 s
```
where it was significantly faster with this patch. I didn't get a measurement for before, because it would take too long (on the order of hours).

parisiale:
👍

PR #121, opened 2016-06-15
(PCP-470) backport PCP-448: Use ConcurrentHashMap for tracking connections

MikaelSmith:
Many Jetty threads can result in modifying the connections or uri map.
When many updates occur concurrently, we can see lots of failed atom
updates resulting in much slower processing.

Switch to using a ConcurrentHashMap. It appears to scale better, and is
at least as fast for smaller numbers of clients.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.6.2
+
+This is a bugfix release
+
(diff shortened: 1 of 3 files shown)


parisiale:
👍 
Same as #120.

PR #122, opened 2016-07-01
(PCP-467) Disable schema validations by default

mruzicka:
In this PR we disable schema validations by default. We achieve that by removing the `^:always-validate` attribute from function definitions in this project.
To compensate for that a new lein alias - `tkv` - is added which is identical to the `tk` alias except it enables the schema validations before it runs trapperkeeper. Additionally the tests are updated to enable the validations by using the `schema.test/validate-schemas` fixture.

### DIFF
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -1,9 +1,12 @@
 (ns user
   (:require [puppetlabs.trapperkeeper.bootstrap :as tk-bootstrap]
             [puppetlabs.trapperkeeper.config :as tk-config]
+            [puppetlabs.trapperkeeper.main :as tk-main]
             [puppetlabs.trapperkeeper.core :as tk]
             [puppetlabs.trapperkeeper.app :as tk-app]
-            [clojure.tools.namespace.repl :as repl]))
+            [clojure.tools.namespace.repl :as repl]
+            [clojure.tools.logging :as log]
+            [schema.core :as s]))
 
 ;; NOTE: in some other projects, where we need to support per-developer config
 ;;  variations, we'll put this code into a namespace called 'user-repl', and
@@ -32,6 +35,13 @@
 (defn go []
   (init)
   (start))
+
+(defn -main
+  "Run trapperkeeper with schema validations enabled"
+  [& args]
+  (s/with-fn-validation
+    (log/info "Running trapperkeeper with schema validations enabled.")
+    (apply tk-main/-main args)))
 
 (defn reset []
   (stop)
--- a/README.md
+++ b/README.md
@@ -14,10 +14,20 @@ And then see [these notes on configuring](doc/configuration.md)
 ## Running the server
 
 For development purposes you can run a broker out of a checkout using
-the *insecure* certificates provided in test-resources/ with the
-following command:
+the *insecure* certificates provided in test-resources/ with either
+the following command:
 
     lein tk
+
+(This one runs the broker with schema validations _disabled_, i.e. the
+same as in production.)
+
+Or with:
+
+    lein tkv
+
+(This one runs the broker with schema validations _enabled_ for greater
+scrutiny.)
 
 ## Documentation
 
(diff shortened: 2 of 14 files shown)


parisiale:
@james-stocks I think it will be nice to update the tests to enable schema validations by using `tkv`

mruzicka:
@parisiale @james-stocks The tests defined in the project don't need the `tkv`, they enable the validations by using the `schema.test/validate-schemas` fixture - e.g. [here](https://github.com/mruzicka/pcp-broker/blob/29b89780a14e77808d548438e6e48dffc5917ad3/test/integration/puppetlabs/pcp/broker/service_test.clj#L68).

dev/user.clj
```diff
@@ -33,6 +36,13 @@
   (init)
   (start))
 
+(defn -main
+  "Run trapperkeeper with schema validations enabled"
```
MikaelSmith: So this overwrites the default that would be in the jar?

MikaelSmith: Is there a way a user could enable this if they want to have more error handling? It'd be nice to enable it in our acceptance testing as well.

MikaelSmith:
@james-stocks were you talking about enabling it in acceptance testing?

dev/user.clj
MikaelSmith: A note in the README would be nice.

mruzicka: Default is no validation. By removing the `^:always-validate` attributes we are effectively disabling it. This is to be able to run the broker with the validations enabled - i.e. the same as before. Actually even more strict than before because it enables the validations for _all_ code including libraries, not just the pcp-broker.

Simple test:
* apply
    ```patch
    --- a/resources/ext/config/conf.d/metrics.conf
    +++ b/resources/ext/config/conf.d/metrics.conf
    @@ -4,7 +4,7 @@ metrics: {
         
         # a server id that will be used as part of the namespace for metrics produced
         # by this server
    -    server-id: localhost
    +    # server-id: localhost
     
         # this section is used to enable/disable JMX reporting
         reporters: {
    ```
* run:
    ```
    lein tk
    ```
    
    which should work
* then run:
    ```
    lein tkv
    ```
    
    which should crash because of a failed validation outside of the pcp-broker:
    ```
    clojure.lang.ExceptionInfo: Input to initialize does not match schema: [(named {:server-id missing-required-key} config)]
    	at puppetlabs.trapperkeeper.services.metrics.metrics_core$eval32435$initialize__32440.invoke(metrics_core.clj:36) ~[na:na]
    ```

parisiale:
@mruzicka I was referring to our acceptance tests...

mruzicka:
@parisiale I was not sure what you meant. For acceptance tests it makes sense.

dev/user.clj
mruzicka: @MikaelSmith the `tkv` alias is meant for the user to be able to run the broker with the validations enabled.
I'm going to update the README.

MikaelSmith: That's actually an interesting effect... but that's not something we had enabled anyway is it.

I think I wasn't clear in some of my questions though. I know the default will now be no validation, and you can run with validation enabled using `lein tkv` (which I'd like to use for acceptance testing in pxp-agent, and wherever else we add tests). Looking at the prismatic schema docs, I don't see a way to enable validation when all you have is the jar.

I'd like to see documentation in the README about `lein tkv`.

mruzicka: Hmm, I'm not sure about your concern: You can enable the validations even when you only have a jar - actually the example I gave above proves that: It did enable validations for the trapperkeeper metrics service despite I only have trapperkeeper-metrics-0.1.1.jar in my local maven repo.

mruzicka: All you need to do is to write a wrapper similar to the [one I added here](https://github.com/puppetlabs/pcp-broker/blob/66e3f0e76eec9491b599d0d58912ce83cec27a04/dev/user.clj#L39-L44).

MikaelSmith: Ok. Could be useful for debugging live deployments.

mruzicka:
@MikaelSmith I've updated the README. It should be now good to merge.

PR #123, opened 2016-07-07
(PCP-485) Improve inventory response performance

MikaelSmith:
Gathering inventory responses for multiple query patterns is slow; for
1000 patterns, the response took ~5 seconds. In that particular case,
find-clients was evaluated lazily, and didn't happen until after
creating the response message and setting its expiry time. This resulted
in the message expiring before it was even sent when find-clients took
several seconds.

Force evaluation of find-clients before constructing the response
message, and speed up find-clients for multiple patterns.

### DIFF
--- a/src/puppetlabs/pcp/broker/in_memory_inventory.clj
+++ b/src/puppetlabs/pcp/broker/in_memory_inventory.clj
@@ -1,10 +1,10 @@
 (ns puppetlabs.pcp.broker.in-memory-inventory
   (:require [clojure.tools.logging :as log]
-            [puppetlabs.pcp.protocol :refer [explode-uri]]
+            [puppetlabs.pcp.protocol :refer [explode-uri uri-wildcard?]]
             [schema.core :as s]))
 
 (defn endpoint-pattern-match?
-  "does an endpoint pattern match the subject value.  Here is where wildards happen"
+  "Does an endpoint pattern match the subject value. Here is where wildcards happen"
   [pattern subject]
   (let [[pattern-client pattern-type] (explode-uri pattern)
         [subject-client subject-type] (explode-uri subject)]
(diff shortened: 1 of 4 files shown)


src/puppetlabs/pcp/broker/in_memory_inventory.clj
```diff
@@ -18,9 +18,11 @@
...
 (defn find-clients
   [inventory patterns]
-  (flatten (map (fn [pattern]
-                  (filter (partial endpoint-pattern-match? pattern) @inventory))
-                patterns)))
+  (let [explicit (set (filter (complement uri-wildcard?) patterns))
+        wildcard (set (filter uri-wildcard? patterns))
+        explicit-matched (clojure.set/intersection explicit @inventory)
+        wildcard-matched (flatten (map #(filter (partial endpoint-pattern-match? %) @inventory) wildcard))]
+    (sort (clojure.set/union explicit-matched (set wildcard-matched)))))
```
parisiale: Wouldn't it be better to remove possible duplicates from `explicit-matched` and `wildcard-matched` by doing `(set (union x y))` instead of `(union x (set y))`?

MikaelSmith: I'm not sure... `explicit-matched` is already a set, so presumably doing `(set wildcard-matched)` will be faster than set of the union. Deduplicating `wildcard-matched` (via `set`) appears to be needed or `union` behaves strangely.
```
user=> (clojure.set/union #{1 2 3} '(1 2 4))
#{1 4 3 2}
user=> (clojure.set/union #{1 2 3} '(1 2 4 2))
(2 3 1 1 2 4 2)
```

The result of `union` is supposed to be a set (the above suggests maybe it's a bit buggy if the inputs are not sets). I sort it to preserve the existing order, which appears to be sorted.

parisiale: Yeah, that's correct.

parisiale:
👍

PR #124, opened 2016-07-08
(PCP-384) Add `/pcp-broker/connect` authz check for association

MikaelSmith:
Adds a `/pcp-broker/connect` authorization check via tk-auth for
association requests.

Also changes tests to use shorter timeouts when expecting no response.

### DIFF
--- a/doc/authorization.md
+++ b/doc/authorization.md
@@ -26,10 +26,12 @@ This would be transformed into the following ring request:
      :form-params {}
      :query-params {"message_type" "http://puppetlabs.com/rpc_blocking_request"
                     "sender" "pcp://client01.example.com/ruby-pcp-client-2251"
-                    "targets" "pcp://client02.example.com/agent"}
+                    "targets" "pcp://client02.example.com/agent"
+                    "destination_report" false}
      :params {"message_type" "http://puppetlabs.com/rpc_blocking_request"
               "sender" "pcp://client01.example.com/ruby-pcp-client-2251"
-              "targets" "pcp://client02.example.com/agent"}}
+              "targets" "pcp://client02.example.com/agent"
+              "destination_report" false}}
 
 And then this can be matched by trapperkeeper-authorization with the following `authorization.conf`.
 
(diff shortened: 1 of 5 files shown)


MikaelSmith:
@mruzicka updated with `associate_request` and `inventory_request` mapping to `/pcp-broker/send` and using authorization.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -287,60 +291,65 @@
...
+      (let [uris (doall (filter (partial get-websocket broker) ((:find-clients broker) (:query data))))
+            response-data {:uris uris}
+            response (-> (message/make-message :message_type "http://puppetlabs.com/inventory_response"
+                                               :targets [(:sender message)]
+                                               :in-reply-to (:id message)
+                                               :sender "pcp:///server")
+                         (message/set-expiry 3 :seconds)
+                         (message/set-json-data response-data))]
+        (s/validate p/InventoryResponse response-data)
+        (accept-message-for-delivery broker (capsule/wrap response)))))
```
parisiale: You'll end up calling `authorized?` again.

parisiale:
I'm not sure we should be invoking `authorized?` in `accept-message-for-delivery`.

doc/authorization.md
```diff
@@ -69,3 +69,110 @@ authorization: {
...
+
+  ]
+}
+```
+
+### Inventory Request
+
+Not all nodes need or should have access to the full inventory of nodes connected to a PCP broker.
+Inventory requests are one way of acquiring that information; another functionally equivalent way
+to discover all connected nodes is a message using a wildcard to specify the target.
```
parisiale: Do you mean, send a multicast message and use the endpoint responses to determine the inventory?
Why don't you mention the `destination_report` flag?

doc/authorization.md
```diff
@@ -69,3 +69,110 @@ authorization: {
...
+
+Both types of requests can be matched by trapperkeeper-authorization with the following
```
parisiale: You have inventory requests. You have PXP requests. But, in general, PCP messages are not requests.

parisiale:
More in general, can we only authorize messages at reception?

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: It calls `authorized?` on the response, where-as above calls it on the request. `authorized?` and `accept-message-for-delivery` get a bit entangled.

MikaelSmith:
`accept-message-for-delivery` authorizes messages because they're being forwarded to other nodes. We could move that. I'm not sure about whether we should authorize other delivery, I'll think about it.

parisiale:
👍

PR #125, opened 2016-07-12
(PCP-487) Increase PCP message expiry in tests

mruzicka:
The tests were prone to hitting the expiry limit, especially on slower HW. By increasing the limit we are increasing the chances the tests succeed on such HW.


parisiale:
Please update the commit and PR titles (maint --> PCP-487).
👍

PR #126, opened 2016-07-13
(PCP-489) Cleanup authorization checks

MikaelSmith:
@parisiale this is as far as I got. I think it should work but for some reason it gets stuck in `certificate-must-match-test`. Changes:
- refactored lots of code
- moved to only authorizing incoming messages
- moved expiration checks to after authorization
- sends `message not authorized` error if authorization check fails


parisiale:
This is solved after splitting the athentication / authorization checks and logging access.
I'll dig deeper to identify a possible data race.

PR #127, opened 2016-07-14
(PCP-496) Set inventory_response message expiration last

MikaelSmith:
Without this change, serializing a list of clients to json can cause the
message to expire before it's sent if that serialization takes a
significant amount of time. Ensure setting the expiration time is the
last thing we do before attempting to send the message.


parisiale:
👍

PR #129, opened 2016-07-20
(PCP-506) Make vNext route truly optional

MikaelSmith:
The vNext route was supposed to be optional, but in practice was not.
Make it truly optional.


parisiale:
👍

PR #133, opened 2016-07-22
(PCP-487) Increase timeouts in tests seriously

mruzicka:
This commit increases expiry timeouts on PCP messages exchanged during tests by a factor of 60. This is to avoid occasional failures in CI we keep seeing.
In contrast to previous attempts to do the same this commit doesn't change the actual timeouts but rather redefines the function which applies the timeout values to the PCP messages: `puppetlabs.pcp.message/set-expiry`.
This has the desired effect of raising the timeouts for both the test code and the tested code at the same time (previous attempts were only raising the timeouts in the test code).
Additionally, timeout is also raised in the `puppetlabs.pcp.testutils.client/recv!` function which, too, was causing test failures by not waiting long enough for messages to be received.
Finally the `it-closes-connections-when-not-running-test` test is redesigned such that it avoids another instance of not waiting long enough for the broker to respond.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -300,9 +300,13 @@
             id (:id request)
             encode (get-in connection [:codec :encode])]
         (if (capsule/expired? capsule)
-          (let [response (make-ttl_expired-message request)]
-            (websockets-client/send! ws (encode response))
-            connection)
+          (do
+            (sl/maplog :trace (assoc (capsule/summarize capsule)
+                                :type :message-expired)
+                       (i18n/trs "Association request '{messageid}' from '{source}' has expired. Sending a ttl_expired."))
+            (let [response (make-ttl_expired-message request)]
+              (websockets-client/send! ws (encode response))
+              connection))
           (let [uri (:sender request)
                 reason (reason-to-deny-association broker connection uri)
                 response (if reason {:id id :success false :reason reason} {:id id :success true})]
--- a/test/integration/puppetlabs/pcp/broker/service_test.clj
+++ b/test/integration/puppetlabs/pcp/broker/service_test.clj
@@ -66,6 +66,20 @@
   (f))
 
 (use-fixtures :once st/validate-schemas)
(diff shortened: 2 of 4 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -300,9 +300,13 @@
             id (:id request)
             encode (get-in connection [:codec :encode])]
         (if (capsule/expired? capsule)
-          (let [response (make-ttl_expired-message request)]
-            (websockets-client/send! ws (encode response))
-            connection)
+          (do
+            (sl/maplog :trace (assoc (capsule/summarize capsule)
```
MikaelSmith: You could avoid the `(do)` by putting logging in the `(let)`. I won't block merging on that though.

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -130,32 +144,44 @@
...
+        ;;    :refused (connection refused)
+        ;;    1011     (broker not started)
+        ;;    1000     (closed because client initated it)
+        ;;    1011     (broker stopping)
+        ;;    :refused (connection refused)
+        ;; though as some of these states may be missed due to timing we test for
+        ;; membership of the set of valid sequences.  If we have more
+        ;; tests like this it might be worth using ztellman/automat to
+        ;; match with a FSM rather than hand-generation of cases.
+        (println @close-codes)
```
MikaelSmith: Intentionally keeping this println?

MikaelSmith:
Two easy questions, overall this looks fine 👍

test/integration/puppetlabs/pcp/broker/service_test.clj
mruzicka: good catch!
that's an omission

src/puppetlabs/pcp/broker/core.clj
mruzicka: I actually had it that way, but the thing I din't like about it was to the call to the `make-ttl_expired-message` before logging the message.

PR #134, opened 2016-07-26
[WIP] (PCP-489) Refactor and improve onMessage handling

parisiale:
Still WORK IN PROGRESS as:
 - [x] I must update the integration tests;-
 - [x] I may add new tests to core_test (PCP-523);
 - [ ] there are a number of TODOs to be resolved;
 - [ ] I will try to factor out PCP-385 changes and split the commit in
   different pieces.

Checking the validity of incoming messages only once, meaning:
 - ignore messages other than associate_request before Association is
   complete;
 - authentication;
 - authorization;
 - TTL check.

Removing the transitions map that keept track of the state of the
association during the message double dispatch mechanism in favour of
making explicit the processing workflow.

Avoiding to update the 'connections' map for each processing message. In
practice, the onMessage handler will update the map only if the
Connection instance being processed is actually updated; that happens
only in case of a successful associate_request.

In case of invalid (as above) associate_request, an unsuccessful
associate_response is sent back to the requester and the WebSocket
connection is closed. That's a behaviour change, as:
 - in case of authentication failure, the broker used to send a PCP
   error and close the WebSocket;
 - in case of authorization failure, the broker used to only close the
   WebSocket.

### DIFF
--- a/locales/messages.pot
+++ b/locales/messages.pot
@@ -85,11 +85,15 @@ msgid ""
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
-msgid "session already associated"
+msgid "Session already associated"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
-msgid "Invalid associate_request ({reason}); closing '{uri}' WebSocket"
+msgid "Replying to '{requester}' with associate_response: '{rawmsg}'"
+msgstr ""
+
+#: src/puppetlabs/pcp/broker/core.clj
+msgid "Invalid associate_request ('{reason}'); closing '{uri}' WebSocket"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
--- a/locales/messages.pot
+++ b/locales/messages.pot
@@ -106,9 +106,8 @@ msgid ""
 "'{remoteaddress}'"
 msgstr ""
 
-#. TODO(ale): is 'superceded' correct? 'superseded'?
 #: src/puppetlabs/pcp/broker/core.clj
-msgid "superceded"
+msgid "superseded"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
--- a/locales/messages.pot
+++ b/locales/messages.pot
@@ -45,16 +45,11 @@ msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
 msgid ""
-"Authorizing '{messageid}' for '{destination}' - '{allowed}': '{message}'"
-msgstr ""
-
-#: src/puppetlabs/pcp/broker/core.clj
-msgid ""
 "Message '{messageid}' for '{destination}' has expired. Sending a ttl_expired."
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
-msgid "Server generated message expired.  Dropping"
+msgid "Server generated message expired; dropping it"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
@@ -86,7 +81,7 @@ msgstr ""
 #: src/puppetlabs/pcp/broker/core.clj
 msgid ""
 "Received session association for '{uri}' from '{commonname}' "
-"'{remoteaddress}'.  Session was already associated as '{existinguri}'"
+"'{remoteaddress}'. Session was already associated as '{existinguri}'"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
@@ -94,9 +89,7 @@ msgid "session already associated"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
-msgid ""
-"Association request '{messageid}' from '{source}' has expired. Sending a "
-"ttl_expired."
+msgid "Invalid associate_request ({reason}); closing '{uri}' WebSocket"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
@@ -105,10 +98,11 @@ msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
 msgid ""
-"Node with uri '{uri}' already associated with connection '{commonname}' "
+"Node with URI '{uri}' already associated with connection '{commonname}' "
 "'{remoteaddress}'"
 msgstr ""
 
+#. TODO(ale): is 'superceded' correct? 'superseded'?
 #: src/puppetlabs/pcp/broker/core.clj
 msgid "superceded"
 msgstr ""
@@ -117,6 +111,38 @@ msgstr ""
 msgid ""
 "Unhandled message type '{messagetype}' received from '{commonname}' "
 "'{remoteaddr}'"
+msgstr ""
+
+#: src/puppetlabs/pcp/broker/core.clj
+msgid ""
+"Authorizing '{messageid}' for '{destination}' - '{allowed}': '{auth-message}'"
+msgstr ""
+
+#: src/puppetlabs/pcp/broker/core.clj
+msgid "Processing PCP message: '{rawmsg}'"
+msgstr ""
+
+#: src/puppetlabs/pcp/broker/core.clj
+msgid "Message not authenticated"
+msgstr ""
+
+#: src/puppetlabs/pcp/broker/core.clj
+msgid "Message not authorized"
+msgstr ""
+
+#. default case
+#: src/puppetlabs/pcp/broker/core.clj
+msgid "unexpected message validation outcome"
+msgstr ""
+
+#: src/puppetlabs/pcp/broker/core.clj
+msgid "Error {0} handling message: {1}"
+msgstr ""
+
+#. TODO(richardc): this could use a different message_type to
+#. indicate an encoding error rather than a processing error
+#: src/puppetlabs/pcp/broker/core.clj
+msgid "Could not decode message"
 msgstr ""
 
 #: src/puppetlabs/pcp/broker/core.clj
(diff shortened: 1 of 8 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,216 +252,319 @@
...
+                        (message/set-json-data response-data)
+                        (message/set-expiry 3 :seconds))]
+        (websockets-client/send! ws (encode message)))
+      (if reason-to-deny
+        (do
+          (sl/maplog
+            :debug {:type   :connection-association-failed
+                    :uri    requester-uri
+                    :reason reason-to-deny}
+            (i18n/trs "Invalid associate_request ({reason}); closing '{uri}' WebSocket"))
```
parisiale: Missing single quotes for `reason`.

test/utils/puppetlabs/pcp/testutils/client.clj
```diff
@@ -60,9 +60,9 @@
                                             :open  (fn [ws]
                                                      (http/send ws :byte (message/encode association-request)))
                                             :byte  (fn [ws msg]
-                                                     (>!! message-chan (message/decode msg)))
+                                                     (put! message-chan (message/decode msg)))
```
parisiale: The blocking '>!' was effectively block those tests that make the broker reply with a sequence of unsuccessful 'associate_response' / WebSocket close.

MikaelSmith:
Success!

src/puppetlabs/pcp/broker/connection.clj
```diff
@@ -45,17 +46,22 @@
...
-                    :created-at (ks/timestamp)}))
+  ;; NB: the 'map->...' constructor comes from schema.core's defrecord
+  (map->Connection
+    {:state          :open
+     :websocket      websocket
+     :codec          codec
+     ;; http://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getHostAddress%28%29
+     :remote-address (try+
+                       (second
+                         (str/split (.. (websockets-client/remote-addr websocket) (toString)) #"/"))
```
MikaelSmith: What was the problem with the prior implementation using getSession?

locales/messages.pot
```diff
@@ -105,10 +102,11 @@ msgstr ""
...
 #: src/puppetlabs/pcp/broker/core.clj
 msgid ""
-"Node with uri '{uri}' already associated with connection '{commonname}' "
+"Node with URI '{uri}' already associated with connection '{commonname}' "
 "'{remoteaddress}'"
 msgstr ""
 
+#. TODO(ale): is 'superceded' correct? 'superseded'?
 #: src/puppetlabs/pcp/broker/core.clj
 msgid "superceded"
```
MikaelSmith: superseded is more correct (though both are accepted in common use)

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -146,6 +109,7 @@
   (let [sender (:sender message)
         in-reply-to (:id message)
         response_data (->> {:id in-reply-to}
+                           ;; TODO(ale): consider removing s/validate; we create it, why validate?
```
MikaelSmith: I'd remove it. If we really want checking, you could make a helper function like
```
(s/defn make-ttl_expired-envelope :- p/TTLExpiredMessage
  [in-reply-to :- s/Str]
  {:id in-reply-to})
```

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -209,6 +177,7 @@
   (when (:destination_report message)
     (let [report {:id (:id message)
                   :targets targets}]
+      ;; TODO(ale): consider removing s/validate; we create it, why validate?
```
MikaelSmith: Same as above, I'd make a helper function
```
(s/defn make-destination-report :- p/DestinationReport
  [id targets]
  {:id id
   :targets targets})
```

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,216 +252,323 @@
...
+        (sl/maplog
+          :debug (assoc (connection/summarize connection)
+                        :uri as
+                        :existinguri uri
+                        :type :connection-already-associated)
+          (i18n/trs "Received session association for '{uri}' from '{commonname}' '{remoteaddress}'. Session was already associated as '{existinguri}'"))
+        (i18n/trs "Session already associated")))))
+
+(s/defn process-associate-request! :- (s/maybe Connection)
+  "Send an associate_response that will be successfull if:
```
MikaelSmith: Should be `successful`.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,216 +252,323 @@
...
+    - a reason-to-deny is not specified as an argument nor determined by
+      reason-to-deny-association;
+    - the requester `client_type` is not `server`;
+    - the specified WebSocket connection has not been associated previously.
+  If the request gets denied, the WebSocket connection will be closed and the
+  function returns nil.
```
MikaelSmith: If the connection is closed, should we remove it from the connection and uri maps?

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,216 +252,323 @@
...
-      "http://puppetlabs.com/inventory_request" (process-inventory-message broker capsule connection)
+      "http://puppetlabs.com/associate_request" (process-associate-request! broker capsule connection)
+      "http://puppetlabs.com/inventory_request" (process-inventory-request broker capsule connection)
       (do
         (sl/maplog :debug (assoc (connection/summarize connection)
                                  :messagetype message-type
                                  :type :broker-unhandled-message)
-                   (i18n/trs "Unhandled message type '{messagetype}' received from '{commonname}' '{remoteaddr}'"))))))
+                   (i18n/trs "Unhandled message type '{messagetype}' received from '{commonname}' '{remoteaddr}'"))
+        connection))))
```
MikaelSmith: Why does this return connection? It doesn't make changes in this case.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,216 +252,323 @@
...
+;; NB(ale): using (s/Maybe Websocket) in the signature for the sake of testing
+(s/defn authorized? :- s/Bool
+  "Check if the message is authorized"
+  [broker :- Broker pcp-message :- Message websocket :- (s/maybe Websocket)]
+  (let [ring-request (make-ring-request pcp-message websocket)
+        {:keys [authorization-check]} broker
+        {:keys [authorized message]} (authorization-check ring-request)
+        allowed (boolean authorized)]
+    (sl/maplog
+      :trace {:messageid    (:id pcp-message)
```
MikaelSmith: I know I wrote this, but I've noticed we do `(capsule/wrap message)` pretty liberally, so that could be done here as well to simplify the maplog call.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,216 +252,323 @@
...
+          :not-authorized
+          :expired
+          :to-be-processed))
+
+(s/defn validate-message :- MessageValidationOutcome
+  "Determine whether the specified message should be processed by checking,
+   in order, if the message: 1) is an associate-request as expected during
+   Session Association; 2) is authenticated; 3) is authorized; 4) expired"
+  [broker :- Broker capsule :- Capsule connection :- Connection is-association-request :- s/Bool]
+  (if (and (= :open (:state connection)) (not is-association-request))
```
MikaelSmith: Could this be a `cond` statement instead? Nested ifs are always ugly.

MikaelSmith:
Overall this seems to still work with all our clients.

src/puppetlabs/pcp/broker/connection.clj
parisiale: Two things here. We have the 'getSession' - 'getRemoteAddress' exposed in `trapperkeeper-webserver-jetty9`, so we shouldn't be going directly to jetty.

As for the parsing, the intent here is to see if this makes sense and later move upstream; 'getRemoteAddress' returns an 'InetSocketAddress' that has a 'toString' method that ends up calling the 'InetAddress' parent 'toString' that, in turn, returns a string in the 'hostname / literal IP address' format; since Jetty does not resolve the hostname, the string becomes something like '/0:0:0:0:0:0:0:1:56824'.

Another way to solve this, perhaps more correct if we were using Java directly :trollface: , would be calling 'InetAddress' 'getAddress' directly as suggested in http://stackoverflow.com/questions/6932902/apache-mina-how-to-get-the-ip-from-a-connected-client

src/puppetlabs/pcp/broker/core.clj
parisiale: Yes, that is wrong.

src/puppetlabs/pcp/broker/core.clj
parisiale: Oh, it was screaming for it.

src/puppetlabs/pcp/broker/core.clj
parisiale: Ok. I'll change the signature to directly pass the capsule then use `(capsule/summarize capsule)`.

src/puppetlabs/pcp/broker/core.clj
parisiale: The current and previous behaviour consists of updating the broker state with `remove-connection!` only by the onClose handler. I think that's correct.

By the way, I've added a comment about that in `core-test/process-associate-request!-test`; perhaps we should document that here.

parisiale:
Updated.

I don't think we want to update the broker's maps other than in the onClose handler.

I've added the make-data... functions for all the places where the broker creates messages, so to enforce the signatures schemas only during tests. But I want to make sure that we're actually validating the envelope data content of each incoming message, since we dropped all schema checks, so I've added a TODO.

Also, before splitting this in multiple commits, I'd like to fix and merge PCP-524, if you agree that 1) it should be fixed and 2) we need that now.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,216 +256,321 @@
...
+(s/defn process-message! :- (s/maybe Connection)
+  "Deserialize, validate (authentication, authorization, and expiration), and
+  process the specified raw message. Return the 'Connection' object associated
+  to the specified 'Websocket' in case it gets modified (hence the '!' in the
+  function name), otherwise nil.
+  Also, log the message validation outcome via 'pcp-access' logger."
+  [broker :- Broker bytes :- message/ByteArray ws :- (s/maybe Websocket)]
+  ;; TODO(ale): since we don't validate responses anymore and the schema
+  ;; checks are not performed in production, ensure that we validate incoming
+  ;; messages (message format and envelope --> in clj-pcp-common)
```
parisiale: That's what I want to do before moving this to a new PR.

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -577,3 +608,5 @@
 (deftest no-vnext-test
   (with-app-with-config app broker-services no-vnext-config
     (is true)))
+
+(deftest sending-an-invalid-message)
```
parisiale: That is wrong.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: I had a reason that wouldn't work. Be interesting to see what you come up with. Might have been we just needed to create a capsule earlier.

src/puppetlabs/pcp/broker/core.clj
parisiale: Created PCP-525/6/7 for this.

parisiale:
Core unit test failing due to missing schema validation; will fix this in PCP-525.

I'm done with this PR; will split changes in different commits.

PR #137, opened 2016-07-29
(PCP-526) Don't do schema validation in tests by default

parisiale:
Adding new profiles that extend, respectively, the dev and test
profiles by using an :injection entry to enable schema validation
via (schema.core/set-fn-validation! true).

Removing the (schema.core/set-fn-validation! true) calls in the test
source files as we now rely on the test-schema-validation lein profile
to enable schema validation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,6 @@ lein: lein2
 jdk:
   - oraclejdk7
   - openjdk7
-script: lein2 test
+script: lein2 with-profile test:test-schema-validation test
 notifications:
   email: false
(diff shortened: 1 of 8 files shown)


parisiale:
I've added an unit test to ensure it works.

parisiale:
Ah! The new test succeed. That means that the PR changes are good.
But we have two failures. I will investigate that.

parisiale:
Updated. The unit test for checking the new test-schema-validation profile will be removed after checking its outcome once more.

parisiale:
I had to explicitly define the test profile to include the needed dependencies.

parisiale:
The new test fails as expected. Other tests pass.
https://gist.github.com/parisiale/c321205d4687fac65a6bbd9b1936adb7

parisiale:
Tests become unstable again and fail (without displaying the reason) when invoked with the `test-schema-verification` profile.
I suspect that using that the proposed approach (using an `:injections` in project.clj to invoke `(schema.core/set-fn-validation! true)`) introduces more overhead (does more verification?) compared with the old one (invoking the schema setter in each test file).

parisiale:
I will close this PR, in order to reference to these failures within PCP-527 (enabling schema verification in CI), and make a new one without Travis changes.

PR #138, opened 2016-08-01
(PCP-526) Don't do schema validation in tests by default

parisiale:
Adding new profiles that extend, respectively, the dev and test
profiles by using an :injection entry to enable schema validation
via (schema.core/set-fn-validation! true).

Refactoring dependencies of test profiles into new `test-base` profile.

Adding new `test-all` task that executes all tests with and without
schema verification.

==

Don't enable schema validation in test sources.

==

After disabling schema validation, capsule/wrap does not enforce that
the Message argument is a valid Message, so we can't test that anymore.

Note that with PCP-525 we will ensure that only valid PCP messages are
accepted by pcp-broker; we can safely drop the Message validation when
creating a Capsule.

==

Schema validation has been disabled by default; don't rely on it when
testing core/determine-next-state behaviour.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -102,13 +102,23 @@
                                   ;; Transitive dependency for lein-cloverage and puppetlabs/kitchensink
                                   [org.clojure/tools.cli "0.3.0"]]
                    :plugins [[lein-cloverage "1.0.6" :excludes [org.clojure/clojure org.clojure/tools.cli]]]}
-             :unit {:source-paths ["test/utils" "test-resources"]
-                    :dependencies [[http.async.client ~http-async-client-version :exclusions [org.clojure/clojure]]]
-                    :test-paths ^:replace ["test/unit"]}
-             :integration [:unit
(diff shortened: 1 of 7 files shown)


parisiale:
Compared to #137, we're not enabling the `test-schema-validation` profile in Travis.

MikaelSmith:
Looks like you still need to update .travis.yml to invoke tests with validation. I'd like to see it as 4 separate Travis jobs, {with validation, without validation} x {openjdk, oracle jdk}.

parisiale:
It doesn't work; I ran Travis 3 times and had 5 failures over 6 attempts on open/oracle jdk. See #137 .

So, what I'd like to do is to find a solution for this (change the way validation is enabled or fix the timeouts / races in the code / test) in PCP-527 (we'll have the same problem on CI) and leave this PR as it is, since the PCP-526 ticket is about testing what actually goes in production.

PR #139, opened 2016-08-02
(PCP-489) Improve onMessage handling

parisiale:
Checking the validity of incoming messages only once, meaning:
 - ignore messages other than associate_request before Association is
   complete;
 - authentication;
 - authorization;
 - TTL check.

Removing the transitions map that keept track of the state of the
association during the message double dispatch mechanism in favour of
making explicit the processing workflow.

Avoiding to update the 'connections' map at each processed message. In
practice, the onMessage handler will now update the map only if the
Connection instance being processed is actually updated; that happens
only in case of a successful associate_request.

Ensure incoming messages are valid Message instances by explicitly
validating against the Message schema; this is necessary as we don't
have schema validation enabled on function calls.

Removing schema validation calls for data content created by the broker.
Factoring out functions that create data content objects in order to
enable schema validation only when globally set (with appropriate lein
profile).

In case of invalid (as above) associate_request, an unsuccessful
associate_response is sent back to the requester and the WebSocket
connection is closed. That's a behaviour change, as:
 - in case of authentication failure, the broker used to send a PCP
   error and close the WebSocket;
 - in case of authorization failure, the broker used to only close the
   WebSocket.

Improving documentation.

Updating tests and creating new ones.

Using the async 'put!' call in testutils/client.clj as the previous
'>!!' blocking function was creating a deadlock in those cases where the
test client was expecting both PCP and WebSocket responses from the
broker under test.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -129,6 +129,9 @@
                                 :message message)
                   (i18n/trs "Authorizing '{messageid}' for '{destination}' - '{allowed}': '{message}'"))
        allowed))))
+;;
+;; Message queueing and processing
+;;
 
 ;; message lifecycle
 (s/defn accept-message-for-delivery :- Capsule
@@ -158,9 +161,10 @@
 (s/defn process-expired-message :- Capsule
   "Enqueue a ttl_expired message to the original message sender"
   [broker :- Broker capsule :- Capsule]
-  (sl/maplog :trace (assoc (capsule/summarize capsule)
-                           :type :message-expired)
-             (i18n/trs "Message '{messageid}' for '{destination}' has expired. Sending a ttl_expired."))
+  (sl/maplog
+    :trace (assoc (capsule/summarize capsule)
+                  :type :message-expired)
+    (i18n/trs "Message '{messageid}' for '{destination}' has expired. Sending a ttl_expired."))
   (let [message (:message capsule)
         sender  (:sender message)]
     (if (= "pcp:///server" sender)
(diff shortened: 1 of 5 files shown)


parisiale:
@MikaelSmith making just one big commit as you've already reviewed this and changes are very coupled. But let me know if you think it's worth separating pieces of this.

Compared to what's in #134: 
 - we now validate incoming message by calling `(s/validate Message message)` in `process-message!`;
 - I've removed access log changes for PCP-385.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,220 +260,314 @@
...
+        (try+
+          (case (validate-message broker capsule connection is-association-request)
+            :to-be-ignored-during-association
+            ;; TODO(ale): log access (PCP-385); doing nothing for now
+            (do)
+            :not-authenticated
+            (let [not-authenticated-msg (i18n/trs "Message not authenticated")]
+              (if is-association-request
+                ;; send an unsuccessful associate_response and close the WebSocket
+                (process-associate-request! broker capsule connection not-authenticated-msg)
```
MikaelSmith: This sends both an unsuccessful associate_response and an error message? Why both?

parisiale: Hmmm, no. If the associate_request is not authenticated, the broker will send an unsuccessful associate_response and then close the WebSocket.

I think it makes sense to send the unsuccessful associate_response; we close the Session Association transaction (request - response) and let the client know what went wrong at PCP layer without assuming anything on the layer below. Otherwise, if we just close the WebSocket, it won't make sense to have *success* and *reason* entries in `associate_response`.

MikaelSmith: Sending the associate_response is fine. It's the next line calling (send-error-message) that's confusing me. It will never succeed.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -281,220 +260,314 @@
...
+                (send-error-message message not-authenticated-msg connection)))
+            :not-authorized
+            (let [not-authorized-msg (i18n/trs "Message not authorized")]
+              (if is-association-request
+                ;; send an unsuccessful associate_response and close the WebSocket
+                (process-associate-request! broker capsule connection not-authorized-msg)
+                ;; TODO(ale): use 'unauthorized' in version 2
+                (send-error-message message not-authorized-msg connection)))
```
MikaelSmith: This `(send-error-message)` will never succeed because the websocket is closed.

PR #142, opened 2016-08-03
(PCP-538) Restore initial delivery of expired messages

MikaelSmith:
Restore attempting to deliver expired messages (but continue to drop
them if delivery fails and they've expired). This makes the broker
somewhat resilient to clock skew.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -33,6 +33,7 @@
    :record-client      IFn
    :find-clients       IFn
    :authorization-check IFn
+   :reject-expired-msg s/Bool
    :uri-map            ConcurrentHashMap ;; Mapping of Uri to Websocket, for sending
    :connections        ConcurrentHashMap ;; Mapping of Websocket session to Connection state
    :metrics-registry   Object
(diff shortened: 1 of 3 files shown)


MikaelSmith:
With `?w=1`, the changes are pretty minor.

parisiale:
Don't you want to process incoming expired messages as well?
I mean https://github.com/puppetlabs/pcp-broker/blob/master/src/puppetlabs/pcp/broker/core.clj#L527.

MikaelSmith:
Hmm yeah, that makes sense. I'll keep working on it.

MikaelSmith:
Updated.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -164,19 +165,17 @@
...
+    (process-expired-message broker capsule)
+    (let [retry-delay (retry-delay capsule)
+          capsule     (capsule/add-hop capsule (broker-uri broker) "redelivery")]
+      (sl/maplog
+        :trace (assoc (capsule/summarize capsule)
+                      :type :message-redelivery
+                      :delay retry-delay)
+        (i18n/trs "Scheduling message '{messageid}' to be delivered in '{delay}' seconds"))
+      (time! (:message-queueing (:metrics broker))
+             (activemq/queue-message delivery-queue capsule (mq/delay-property retry-delay :seconds))))))
```
parisiale: Same logic; correctly using the factored out `capsule/expired?`.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -203,7 +202,7 @@
 (s/defn deliver-message
   "Message consumer. Delivers a message to the websocket indicated by the :target field"
   [broker :- Broker capsule :- Capsule]
-  (if (capsule/expired? capsule)
+  (if (and (:reject-expired-msg broker) (capsule/expired? capsule))
```
parisiale: We don't end up here for redelivering an expired message, as `handle-delivery-failure` remains the same. 👍

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -456,7 +455,7 @@
          (not is-association-request)) :to-be-ignored-during-association
     (not (authenticated? capsule connection)) :not-authenticated
     (not (authorized? broker capsule (:websocket connection))) :not-authorized
-    (capsule/expired? capsule) :expired
+    (and (:reject-expired-msg broker) (capsule/expired? capsule)) :expired
```
parisiale: It gets cleaner using the member flag only here. 👍 
We can live without a log message for signalling an expired incoming message; it will be until PCP-187. We can always complicate the workflow a bit and introduce one later if we change our minds.

parisiale:
`lein test-all` passes in my laptop.

parisiale:
Refer to PCP-245 and #112 for previous related changes.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: This change we didn't discuss as much, I'd be ok with it still checking expired now that I've thought it through. It should only trigger on messages that are in the delivery queue too long; those aren't the types of message we're worried about with PCP-538.

PR #144, opened 2016-08-05
(PCP-525) Rely on pcp-common.message for validating Messages

parisiale:
- bumping to clj-pcp-common 0.5.2
- fixing process-message! test
- removing s/validate call for decoded messages in process-message!
- adding more unit tests for process-messasge!


MikaelSmith:
👍

PR #148, opened 2016-09-13
(maint) Minor cleanup for puppet-server

MikaelSmith:
Bumps project dependencies and delays reading the broker certificate until service starts so it can be run in a puppet-server JVM. Also ignore the pcp access log.


MikaelSmith:
Should be superceded by @mruzicka's work on PCP-597.

PR #151, opened 2016-11-11
(PCP-597) Enable puppetserver collocating

mruzicka:
This PR updates the project to use clj-parent for dependency versioning and changes the way how the broker name is determined. Here are the relevant commit descriptions:

### Use clj-parent for dependency versioning
Make use of the `parent-project` and `managed-dependencies` support in Leiningen 2.7.1 and rely on the `puppetlabs/clj-parent` project to define the versions of the majority of the project's dependencies.

### Get broker name from webserver context
Previously the broker name was read from the certificate specified in the ssl configuration of the webserver hosting the broker. This is not possible any more as:
- ssl may not be configured for the webserver at all
- the ssl configuration actually used by the webserver may differ from that specified in its configuration (it can be overridden)

So newly we try to obtain the broker name from the ssl-context actually used by the webserver hosting the broker and if that fails we fall back to using the hostname of the host running the broker.

### DIFF
--- a/src/puppetlabs/pcp/broker/service.clj
+++ b/src/puppetlabs/pcp/broker/service.clj
@@ -3,8 +3,9 @@
             [puppetlabs.pcp.broker.in-memory-inventory :refer [make-inventory record-client find-clients]]
             [puppetlabs.structured-logging.core :as sl]
             [puppetlabs.trapperkeeper.core :as trapperkeeper]
-            [puppetlabs.trapperkeeper.services :refer [service-context]]
+            [puppetlabs.trapperkeeper.services :refer [service-context get-service]]
             [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
+            [puppetlabs.trapperkeeper.services.webserver.jetty9-core :as jetty9-core]
             [puppetlabs.i18n.core :as i18n]))
 
 (trapperkeeper/defservice broker-service
@@ -19,9 +20,6 @@
           accept-consumers   (get-in-config [:pcp-broker :accept-consumers] 4)
           delivery-consumers (get-in-config [:pcp-broker :delivery-consumers] 16)
           inventory          (make-inventory)
-          ssl-cert           (if-let [server (get-server this :v1)]
-                               (get-in-config [:webserver (keyword server) :ssl-cert])
-                               (get-in-config [:webserver :ssl-cert]))
           broker             (core/init {:activemq-spool activemq-spool
                                          :accept-consumers accept-consumers
                                          :delivery-consumers delivery-consumers
@@ -30,8 +28,7 @@
                                          :find-clients   (partial find-clients inventory)
                                          :authorization-check authorization-check
                                          :get-metrics-registry get-metrics-registry
-                                         :get-route (partial get-route this)
-                                         :ssl-cert ssl-cert})]
+                                         :get-route (partial get-route this)})]
       (register-status "broker-service"
                        (status-core/get-artifact-version "puppetlabs" "pcp-broker")
                        1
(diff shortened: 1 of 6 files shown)


src/puppetlabs/pcp/broker/service.clj
```diff
@@ -19,33 +20,37 @@
...
     (sl/maplog :info {:type :broker-start} (i18n/trs "Starting broker service"))
-    (let [broker (:broker (service-context this))]
-      (core/start broker))
-    (sl/maplog :debug {:type :broker-started} (i18n/trs "Broker service started"))
-    context)
+    (let [broker (:broker context)
+          broker-name (or (:broker-name broker)
+                          (some-> (get-service this :WebserverService)
+                                  service-context
+                                  (jetty9-core/get-server-context (get-server this :v1))
```
MikaelSmith: Code above used to check whether `(get-server this :v1)` returned anything. Could this still return nil?

Also, do we need to use the `:v1` argument? The [docs for get-server](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-service.md#get-server) points to a one-argument version, but it's not clear to me what it would do in this case.

src/puppetlabs/pcp/broker/service.clj
```diff
@@ -19,33 +20,37 @@
...
+                                  core/get-webserver-cn)
+                          (core/get-localhost-hostname))
+          broker (assoc broker :broker-name broker-name)
+          context (assoc context :broker broker)]
+      (core/start broker)
+      (sl/maplog :debug {:type :broker-started} (i18n/trs "Broker service started"))
```
MikaelSmith: I think it'd be nice to print the broker name as part of this for debugging.

src/puppetlabs/pcp/broker/service.clj
```diff
@@ -19,33 +20,37 @@
...
-    context)
+    (let [broker (:broker context)
+          broker-name (or (:broker-name broker)
+                          (some-> (get-service this :WebserverService)
+                                  service-context
+                                  (jetty9-core/get-server-context (get-server this :v1))
+                                  core/get-webserver-cn)
+                          (core/get-localhost-hostname))
+          broker (assoc broker :broker-name broker-name)
+          context (assoc context :broker broker)]
```
MikaelSmith: These two lines could instead be `context (assoc-in context [:broker :broker-name] broker-name)`.

If you do that, at line 40 we could just use `(get-in broker [:broker :broker-name])`.

src/puppetlabs/pcp/broker/service.clj
MikaelSmith: I get an error at this line running `lein tk`:
```
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to clojure.lang.IFn
	at puppetlabs.trapperkeeper.services.webserver.jetty9_core$get_server_context.invokeStatic(jetty9_core.clj:926)
	at puppetlabs.trapperkeeper.services.webserver.jetty9_core$get_server_context.invoke(jetty9_core.clj:917)
	at puppetlabs.pcp.broker.service$reify__32608$service_fnk__6004__auto___positional$reify__32623.start(service.clj:43)
```

mruzicka: It's ok for the `(get-server this :v1)` to return `nil` (it does happen if the old-style single server configuration syntax is used) as the `(jetty9-core/get-server-context )` [can handle that](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/1.5.11/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj#L919-L921).

AFAIKT the `:v1` has to be used as the single argument version of the `(get-server )` function only works for service routing definitions with only one endpoint, e.g.:
```
  "puppetlabs.trapperkeeper.services.status.status-service/status-service": {
    route: "/status"
    server: "pcp-broker"
  }
``` 
cf. with:
```
  "puppetlabs.pcp.broker.service/broker-service": {
    v1: {
      route: "/pcp"
      server: "pcp-broker"
    }
    vNext: {
      route: "/pcp/vNext"
      server: "pcp-broker"
    }
  }
```

src/puppetlabs/pcp/broker/service.clj
mruzicka: 👍

src/puppetlabs/pcp/broker/service.clj
MikaelSmith: Yeah, if there's no `v1` route specified, the server fails at https://github.com/puppetlabs/pcp-broker/blob/master/src/puppetlabs/pcp/broker/core.clj#L730.

src/puppetlabs/pcp/broker/service.clj
MikaelSmith:
```
      (sl/maplog :debug {:type :broker-started
                         :brokername broker-name} (i18n/trs "Broker service started at '{brokername}'"))
```
seems to work well.

src/puppetlabs/pcp/broker/service.clj
mruzicka: Yeah, I had been considering that, but since the `broker` is used on line 48:
```clojure
      (core/start broker)
```
I ended up using this verbose syntax.

PR #153, opened 2016-11-18
(maint) Update to clj-parent 0.2.4

nicklewis:
This pins a version of tk-metrics so that orchestrator and pcp-broker
will be compatible without requiring an override.


nicklewis:
This should probably be merged after puppetlabs/pe-orchestration-services#38 and puppetlabs/orchestrator#477 have gone through.

PR #154, opened 2016-12-10
(PCP-645) implement v2 protocol

wkalt:
This has been tested manually against pxp-agent and clj-pxp-puppet, as well as a basic text-based Websocket client. I also ran it in pxp-agent acceptance tests with modest modification to test setup (https://github.com/puppetlabs/pxp-agent/pull/529).

It'll need a squash before merge. It also depends on https://github.com/puppetlabs/clj-pcp-common/pull/32, which needs to be merged and released for unit tests to pass.


src/puppetlabs/pcp/broker/connection.clj
```diff
@@ -43,28 +32,57 @@
...
+        "/pcp/v2.0" :v2
+        "/pcp/vNext" :vNext))))
+
+(s/defn summarize-ws
+  [ws :- Websocket]
+  {:commonname (ws->common-name ws)
+   :remoteaddress (ws->remote-address ws)})
+
+;; TODO this doesn't actually work for supplied client types. figure out what
+;; to do.
```
wkalt: need to fix

test/utils/puppetlabs/pcp/testutils/client.clj
```diff
@@ -25,7 +27,7 @@
     (http/send ws-client :byte bytes))
   (send! [_ message]
     (http/send ws-client :byte (message/encode message)))
-  (recv! [this] (recv! this (* 10 5 1000)))
+  (recv! [this] (recv! this (* 3 1000)))
```
wkalt: this seemed to make the tests run faster, even when they all passed

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -139,17 +162,20 @@
...
+        error-msg (-> (message/make-versioned-message
+                        "v2.0"
+                        {:message_type "http://puppetlabs.com/error_message"
+                         :sender "pcp:///server"
+                         :data data-content
+                         :in_reply_to (:id in-reply-to-message)}))
+        {:keys [websocket]} connection
+        version (connection/ws->protocol-version websocket)
+        encoded-message (encode error-msg version)]
+    (websockets-client/send! websocket encoded-message)
```
MikaelSmith: Links to my comment at https://github.com/puppetlabs/clj-pcp-common/pull/31#discussion_r92051978. I guess I'm ok with this division if it works, just seems unclear to me whether it should work.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -165,27 +191,29 @@
 (s/defn multicast-message?
   "Returns a boolean specifying whether the message uses multicast in the target field."
   [message :- Message]
-  (if-let [targets (:targets message)]
-    (if-let [first-target (first targets)]
-      (or (not= 1 (count targets)) (p/uri-wildcard? first-target)))))
+  (if-let [target (:target message)]
+    (if-let [first-target (first target)]
```
MikaelSmith: Seems like it might not work right if `:target` is a single URI (rather than an array).

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -237,97 +245,95 @@
...
-            :uri requester-uri
-            :state :associated))))))
+  [broker :- Broker
+   request :- Message
+   connection :- Connection]
+  ;; NB(ale): don't validate the associate_request as there's no data chunk...
+  (let [ws (:websocket connection)
+        version (connection/ws->protocol-version ws)
+        id (:id request)
+        requester-uri (:sender request)
```
MikaelSmith: We shouldn't just accept the sender URI, we should generate it from the websocket connection. Your current code allows masquerading as another client. Huh, I was sure this was checked, but now I can't find where.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -237,97 +245,95 @@
...
+                         :rawmsg message}
+                 (i18n/trs "Replying to '{requester}' with associate_response: '{rawmsg}'"))
+      (websockets-client/send! ws (encode message version))
+      (.put inventory requester-uri connection)
+      (assoc connection :uri requester-uri)))
 
 (s/defn make-inventory_response-data-content :- p/InventoryResponse
-  [uris]
-  {:uris uris})
+  [uris broker]
```
MikaelSmith: We should pass in version, rather than broker. We should also make sure querying uris and version are done as one atomic query.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: Ah right, it's `authenticated?`. Nevermind, this code should be safe.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -139,17 +162,20 @@
...
-                                            :sender "pcp:///server")
-                      (message/set-json-data data-content))
-        error-msg (if in-reply-to-message
-                    (assoc error-msg :in-reply-to (:id in-reply-to-message))
-                    error-msg)
-        {:keys [codec websocket]} connection
-        encode (:encode codec)]
-    (websockets-client/send! websocket (encode error-msg))
+        error-msg (-> (message/make-versioned-message
+                        "v2.0"
```
wkalt: right. everything internal is presumed to be on the latest version. It may be worth sticking latest version in a var for maintainability though.

src/puppetlabs/pcp/broker/core.clj
wkalt: yeah, the array was a brain fart. Will go back through and convert.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -497,12 +498,13 @@
...
-          (assoc (connection/summarize connection) :type :deserialization-failure
-                                                   :outcome "DESERIALIZATION_ERROR")
+          (assoc (connection/summarize-ws ws)
+                 :type :deserialization-failure
+                 :outcome "DESERIALIZATION_ERROR")
           "{outcome} {remoteaddress} {commonname} unknown unknown unknown unknown unknown")
         ;; TODO(richardc): this could use a different message_type to
         ;; indicate an encoding error rather than a processing error
-        (send-error-message nil (i18n/trs "Could not decode message") connection)))))
+        (send-error-message nil (i18n/trs "Could not decode message") {:websocket ws}))))
```
MikaelSmith: This looks wrong. Isn't `ws` already the websocket?

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -514,8 +516,7 @@
     (if-not (= :running @(:state broker))
       (websockets-client/close! ws 1011 (i18n/trs "Broker is not running"))
       (when-let [connection (process-message! broker bytes ws)]
-        (assert (instance? Connection connection))
-        (.put (:connections broker) ws connection)))))
+        (assert (instance? Connection connection))))))
```
MikaelSmith: We could probably drop the `when-let` and `assert`.

src/puppetlabs/pcp/broker/connection.clj
```diff
@@ -43,28 +32,57 @@
...
+    (catch Exception _
+      nil)))
+
+(defn ws->protocol-version
+  [ws]
+  (if (= ws :dummy-ws) ;; TODO this is for testing; pretty sloppy
+    :v2
+    (let [path (.. ws getSession getUpgradeRequest getRequestURI getPath)]
+      (case path
+        "/pcp/v1.0" :v1
```
MikaelSmith: I'd rather not hard-code these. They definitely won't match this in PE. Also, drop vNext, it's no longer used.

They should query the webroutes service for the relevant path.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -82,33 +110,28 @@
...
 ;; connection map lifecycle
 (s/defn add-connection! :- Connection
   "Add a Connection to the connections to track a websocket"
-  [broker :- Broker ws :- Websocket codec :- Codec]
-  (let [connection (connection/make-connection ws codec)]
-    (.put (:connections broker) ws connection)
+  [broker :- Broker
+   ws :- Websocket]
+  (let [{:keys [common-name] :as connection} (connection/make-connection ws)
+        path (str "pcp://" common-name "/agent")]
```
MikaelSmith: You have `ws->uri`, we should use that. Also, `agent` is just the default for v1 messages. v2 should extract it from the path (using similar logic to remove the prefix as I noted for `ws->protocol-version`).

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -82,33 +110,28 @@
...
 
 (s/defn remove-connection!
   "Remove tracking of a Connection from the broker by websocket"
-  [broker :- Broker ws :- Websocket]
-  (if-let [uri (get-in (:connections broker) [ws :uri])]
-    (.remove (:uri-map broker) uri))
-  (.remove (:connections broker) ws))
+  [broker :- Broker
+   common-name]
+  (when-let [path (str "pcp://" common-name "/agent")]
```
MikaelSmith: Same note about using `ws->uri` and `agent` only applying to v1 clients.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: `send!` maps to `sendBytes` or `sendText` at https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/src/puppetlabs/trapperkeeper/services/webserver/experimental/jetty9_websockets.clj#L25-L36. I'd really rather send text as text, and have onText expect text (rather than a byte array).

It's ok if onText only works with the v2 endpoint, and onBytes only works with the v1 endpoint.

src/puppetlabs/pcp/broker/core.clj
wkalt: I left this in place because some tests depend on it today. We can probably work it out though.

test/utils/puppetlabs/pcp/testutils/client.clj
```diff
@@ -39,22 +41,22 @@
...
   "Makes a client for testing"
   [& {:keys [certname uri version modify-association check-association modify-association-encoding]
       :or {modify-association identity
            modify-association-encoding identity
            check-association true
-           version "vNext"}}]
-  (let [uri                 (or uri (str "pcp://" certname "/test"))
-        association-request (modify-association (make-association-request uri))
+           version "v2.0"}}]
+  (let [uri                 (or uri (str "pcp://" certname "/agent"))
```
wkalt: on revisiting, I think this may be acceptable for test utils

wkalt:
@MikaelSmith I think I've addressed all your comments. There are updates on clj-pcp-common, and a new PR on trapperkeeper-webserver-jetty9 here: https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/156

I've left the commits split out here in case it's useful to you; I'm assuming they'll be squashed prior to merge. Sorry about anything I've missed and have an excellent holiday!

MikaelSmith:
I have some work on this I'll put up tomorrow. I need to fix a few tests and clean it up.

MikaelSmith:
Still a work in progress. I added a commit based on changes proposed in https://github.com/puppetlabs/clj-pcp-common/pull/32. Also tested with various clients and made fixes. Need to rework tests now, and I'll want to run this through the full pxp-agent test suite before merging as well.

It also doesn't quite match the pcp-specifications definition of v2 yet.

MikaelSmith:
I plan to update the docs folder still.

MikaelSmith:
@puppetlabs/pcp-broker-maintainers I think this is ready for review. I've tested it manually against pxp-agent and clj-pxp-puppet, as well as a basic text-based Websocket client. I also ran it in pxp-agent acceptance tests with modest modification to test setup (https://github.com/puppetlabs/pxp-agent/pull/529).

It'll need a squash before merge. It also depends on https://github.com/puppetlabs/clj-pcp-common/pull/32, which needs to be merged and released for unit tests to pass.

src/puppetlabs/pcp/broker/connection.clj
```diff
@@ -21,6 +22,7 @@
 (s/defrecord Connection
              [websocket :- Websocket
               remote-address :- s/Str
+              codec :- Codec
```
wkalt: I don't understand the reason for tracking this in every connection when the codecs used in practice are so predictable.

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -263,33 +271,35 @@
...
                                  :sender "pcp://client01.example.com/agent"
                                  :data {:query ["pcp://client01.example.com/agent"]}})]
                    (client/send! client request)
                    (let [response (client/recv! client)
-                         data (message/get-json-data response version)]
+                         data (client/get-data response version)]
                      (is (= "http://puppetlabs.com/inventory_response" (:message_type response)))
-                     (is (= (:id request) (:in_reply_to response)))
+                     (if (= version "v1.0")
+                       (is (= (:id request) (:in-reply-to response)))
```
wkalt: we should update the v1 spec to mention in-reply-to

test/unit/puppetlabs/pcp/broker/message_test.clj
```diff
@@ -0,0 +1,40 @@
...
+                 v1-encode
+                 v1-decode)))))
+  (testing "message survives v2 roundtrip"
+    (let [message (message/make-message
+                   {:sender "pcp://gangoffour/entity"
+                    :message_type "ether"
+                    :target "pcp://gangoffour/entity"})]
+      (is (= (-> message
+                 v2-encode
+                 v2-decode))))))
```
wkalt: this test and the one above it will always pass

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -121,21 +121,28 @@
...
-    (is (= nil (:in_reply_to message)))
-    (when check-in-reply-to
-      (is (:in_reply_to message))))
-  (is (= expected-description (:description (message/get-json-data message version)))))
-
-(defn association-success
-  [message version]
-  "Assert that the message is a successful association response."
+  (let [data (client/get-data message version)]
+    (if (= "v1.0" version)
```
wkalt: like you mentioned in the other patch it would be preferable to switch to a keyword representation of version throughout -- not sure if we want to take that step at this point.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -433,27 +400,37 @@
...
    "{accessoutcome} {remoteaddress} {commonname} {source} {messagetype} {messageid} {destination}"))
 
-;; NB(ale): using (s/maybe Websocket) in the signature for the sake of testing
-(s/defn process-message! :- (s/maybe Connection)
+(s/defn fill-optional
+  [message :- Message
+   uri :- p/Uri]
+  (cond-> message
+    (not (:sender message)) (assoc :sender uri)
+    (not (:target message)) (assoc :target "pcp:///server")))
```
wkalt: this could be simplified with the new assoc-when function in utils

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -346,17 +271,19 @@
...
-       (message/make-message
-        latest-protocol-version
-        {:message_type "http://puppetlabs.com/inventory_response"
-         :target (:sender message)
-         :in_reply_to (:id message)
-         :sender "pcp:///server"
-         :data response-data})
-       connection)))
+    (let [requester-uri (:sender message)
+          response-data (make-inventory_response-data-content broker (:query data))
```
wkalt: pretty minor but these underscores bugged me a bit before -- is anyone attached to them?

src/puppetlabs/pcp/broker/message.clj
```diff
@@ -0,0 +1,91 @@
...
+
+(s/defn v1->v2 :- m2/Message
+  "Transform a v1 Message to a v2 Message."
+  [v1-msg :- m1/Message]
+  (let [data' (m1/get-json-data v1-msg)
+        message' (-> v1-msg
+                     (dissoc :expires :destination_report)
+                     validate-not-multicast
+                     (rename-keys {:in-reply-to :in_reply_to})
+                     (rename-keys {:targets :target})
```
wkalt: we can consolidate this line and the one above

src/puppetlabs/pcp/broker/message.clj
```diff
@@ -0,0 +1,91 @@
...
+                     (update-when [:target] first)
+                     (dissoc :_chunks))
```
wkalt: this can be consolidated with line 34

src/puppetlabs/pcp/broker/message.clj
```diff
@@ -0,0 +1,91 @@
...
+                     (when-let [first-target (first target)]
+                       (or (not= 1 (count target))
+                           (p/uri-wildcard? first-target))))]
+    (cond-> message
+      multicast? (assoc :multicast-message true))))
+
+(s/defn v1->v2 :- m2/Message
+  "Transform a v1 Message to a v2 Message."
+  [v1-msg :- m1/Message]
+  (let [data' (m1/get-json-data v1-msg)
```
wkalt: may as well just call this "data"

wkalt: ah, never mind -- I see what you're doing below. Conventionally I'd expect data' to be bound after data.

wkalt: (likewise with message)

src/puppetlabs/pcp/broker/message.clj
```diff
@@ -0,0 +1,91 @@
...
+        message (cond-> message'
+                  (and is-error (:id data')) (assoc :in_reply_to (:id data')))
+        data (cond-> data'
+               is-error (:description))]
+    (cond-> message
+      data (assoc :data data))))
+
+(s/defn v2->v1 :- m1/Message
+  "Transform a v2 Message to a v1 Message."
+  [{:keys [message_type data] :as v2-msg :- m2/Message}]
```
wkalt: this should be
```
[{:keys [message_type data] :as v2-msg} :- m2/Message]
```
for the validation to work

wkalt: may as well destructure in_reply_to here too to use it below

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -237,43 +232,37 @@
...
+  (testing "Unsuccessful associate_response and WebSocket closes if client requests new association"
+    (with-app-with-config app broker-services broker-config
+      (dotestseq [version protocol-versions]
+                 (with-open [client (client/connect :certname "client01.example.com"
+                                                    :uri "pcp://client01.example.com/test"
+                                                    :check-association false
+                                                    :version version)]
+                   (let [pcp-response (client/recv! client)
+                         close-websocket-msg (client/recv! client)]
+                     (is (is-association_response pcp-response version false "Session already associated"))
```
wkalt: the is here will cause redundant failures

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -442,99 +433,85 @@
...
+    (with-open [sender   (client/connect :certname "client01.example.com"
+                                         :version "v1.0")
+                receiver (client/connect :certname "client02.example.com"
+                                         :version "v1.0")]
+      (let [message' (m1/make-message
+                      {:sender "pcp://client01.example.com/agent"
+                       :targets ["pcp://client02.example.com/agent"]
+                       :destination_report true
+                       :message_type "greeting"})
+            message (m1/set-json-data message' "Hello")]
```
wkalt: also seems like the double binding isn't needed here

src/puppetlabs/pcp/broker/message.clj
wkalt: see https://en.wikipedia.org/wiki/Prime_(symbol)#Use_in_mathematics.2C_statistics.2C_and_science for background

test/utils/puppetlabs/pcp/testutils/client.clj
```diff
@@ -24,8 +28,11 @@
   (sendbytes! [_ bytes]
     (http/send ws-client :byte bytes))
   (send! [_ message]
-    (http/send ws-client :byte (message/encode message)))
-  (recv! [this] (recv! this (* 10 5 1000)))
+    (let [version (if (:_chunks message) :v1 :v2)] ;; TODO do something better here
```
wkalt: this TODO may or may not apply -- if presence of :_chunks is sufficient we can remove it.

wkalt:
took a pass through. mostly minor things, and +1 on the renamespacing stuff. Reintroduction of codecs feels like a smell to me, but I never understood why they were there so I may be missing some background.

MikaelSmith:
@wkalt the codec is a mechanism to associate connection with a version, so we don't have to pass it through in all functions that need to know, and removes conditional logic (call a function and pass in the version to get the correct callback). It more directly applies the thing we care about - conversion to a common format - rather than a version that can be associated with a conversion function. If we're worried about memory (8 bytes per connection) I could switch to threading version through all the operations.

test/integration/puppetlabs/pcp/broker/service_test.clj
MikaelSmith: For testing, this is fine.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: They correspond to an actual keyword (`inventory_response`) and I'd rather avoid too many changes in a PR this complex.

MikaelSmith:
Updated.

wkalt:
@MikaelSmith No, I'm not worried about the memory cost. I understand your perspective. The thing I like about version dispatch in encode/decode is that since version is already available in connections via the ws adapter, using it allows the whole codec concept to be dropped with a minor change. It's not a big deal though; this patch seems fine.

MikaelSmith:
@wkalt I don't know that it is. At least, the way we stash path for `ws->uri` doesn't provide the full path, just everything after the route. Possibly the way we were retrieving it before (`getSession getUpgradeRequest getPath`) would allow comparing to the route.

test/unit/puppetlabs/pcp/broker/core_test.clj
```diff
@@ -210,138 +152,101 @@
...
           (let [broker     (make-test-broker)
-                connection (add-connection! broker "ws" identity-codec)
-                connection (process-associate-request! broker message connection)]
+                connection-uri (->> (add-connection! broker :dummy-ws dummy-codec)
+                                    (process-associate-request! broker message)
+                                    :uri)]
             (is (not (realized? @closed)))
-            (is (= :associated (:state connection)))
-            (is (= "pcp://localhost/controller" (:uri connection)))))
-        (testing "Associates a client already associated on a different session, but disconnects the first"
```
MikaelSmith: Actually, rejecting the new one is tricky to do, and I don't see a reason to change from the old behavior. So going with superseding the old one.

src/puppetlabs/pcp/broker/websocket.clj
```diff
@@ -0,0 +1,48 @@
...
+(defn ws->remote-address
+  "Extract IP address and port out of the string representation
+  of the InetAddress instance ('hostname/socket' format), so that we
+  don't end up with a remote-address like '/0:0:0:0:0:0:0:1:56824'. See:
+  http://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getHostAddress%28%29
+  http://stackoverflow.com/questions/6932902/apache-mina-how-to-get-the-ip-from-a-connected-client"
+  [ws]
+  (try+
+   (second
+    (str/split (.. (websockets-client/remote-addr ws) (toString)) #"/"))
```
mruzicka: We should be able to extract the address in a more straightforward way than parsing it from a string using a regexp.

mruzicka: We should be able to extract the address in a more straightforward way than parsing it from a string using a regexp.

mruzicka:
In the end only one minor suggestion survived my long review, and it's a minor one. So let's move forward with this and address it later.

MikaelSmith:
Sounds good, thanks.

src/puppetlabs/pcp/broker/websocket.clj
MikaelSmith: Yeah. I think this was already there (just moved to a new function) so I didn't mess with it. There are ways to construct it out of hostname and port; I looked at that when I was adding to trapperkeeper-webserver-jetty9, but decided not to change it because it'd break existing behavior.

PR #158, opened 2017-01-12
(maint) remove erict from maintainers

er0ck:


MikaelSmith:
Hmm, getting some intermittent failures. I'll have to look into those.

PR #159, opened 2017-01-13
(PCP-600) Inventory updates

mruzicka:
This PR introduces support for inventory updates. Here is an outline of the used data structures and the operation.

## The used data structures:
- `:routing-map` (ConcurrentHashMap) a map of client Uri to websocket connection, used for message routing; the structure supports parallel updates, but it's impossible to obtain a consistent view of it without locking it globally, which is not done so as not to sacrifice performance which means it's not suitable for building the inventory responses (which need to be consistent especially when subscribing for periodic updates)
- `:changes` (ConcurrentLinkedDeque) a queue of changes to the :routing-map (essentially a serialization of the client connection and disconnection events) which serves as a buffer between the parallel capable but inherently inconsistent :routing-map and the :inventory and :updates data structures described below, it can handle parallel updates and for performance reasons is never locked globally either
- `:inventory` (HashSet) a set of the Uris of the clients known to the broker; this structure is populated by applying the changes accumulated in the :changes queue at appropriate times; it can be (and must be) locked globally when its state is modified or read but as a result its consistent view can be obtained
- `:updates` (LinkedList) a queue of the same update events as in the :changes queue, the events are added to it at the same times the :inventory set is updated, it must be locked globally when updated or read but similarly to the :inventory this makes it possible to obtain its consistent view
- `:subscriptions` (HashMap) a map of client Uri to subscription record, which stores the following pieces of information:
  - `:connection`, the websocket connection on which the subscription was received, this is used to detect stale subscriptions
  - `:pattern-sets`, a form of Uri filters of the inventory updates the subscriber is interested in receiving
  - `:next-update`, an offset to the :updates queue of the next update that should be sent to this subscriber (provided that it matches the :pattern-sets)

## The operation:
- when a client connects/disconnects a mapping is added to/removed from the :routing-map structure and an appropriate event record is added to the :changes queue; note that these structures are designed to handle parallel updates so no explicit locking of those is needed
- when an inventory request is received then events accumulated in the :changes queue are applied to the :inventory and :updates structures to obtain their up-to-date yet consistent views, subsequently the inventory response is composed from the :inventory view (or snapshot as it is referred to in the code); if the periodic updates are requested, the client is added to the :subscriptions map, the :next-update in the subscription record is set to the length of the :updates queue (as all the previous updates are already reflected in the inventory snapshot)
- updates are sent periodically to subscribed clients, on each iteration the events accumulated in the :changes queue are applied to the :inventory and :updates structures to obtain their up-to-date yet consistent views, then the view of the :updates queue is used to compose the update messages for each subscribed client, the :next-update offset of each subscription is updated as the corresponding inventory update message is sent, when updates are sent to all subscribed clients the processed records from the :updates queue are removed and the :next-update offsets of all subscriptions are updated accordingly

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -38,7 +38,7 @@
                  ;; try+/throw+
                  [slingshot]
 
-                 [puppetlabs/pcp-common "1.0.0"]
+                 [puppetlabs/pcp-common "1.1.0"]
 
                  [puppetlabs/i18n]]
 
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.pcp.broker.core
-  (:require [metrics.gauges :as gauges]
-            [puppetlabs.experimental.websockets.client :as websockets-client]
+  (:require [puppetlabs.experimental.websockets.client :as websockets-client]
+            [puppetlabs.pcp.broker.shared :as shared :refer
+             [Broker get-connection summarize send-error-message send-message deliver-message deliver-server-message]]
             [puppetlabs.pcp.broker.connection :as connection :refer [Codec]]
             [puppetlabs.pcp.broker.websocket :refer [Websocket ws->uri ws->client-type]]
             [puppetlabs.pcp.broker.metrics :as metrics]
@@ -19,28 +20,10 @@
             [slingshot.slingshot :refer [throw+ try+]]
             [puppetlabs.i18n.core :as i18n])
   (:import [puppetlabs.pcp.broker.connection Connection]
-           [clojure.lang IFn Atom]
+           [clojure.lang IFn]
            [java.util.concurrent ConcurrentHashMap]
            [java.net InetAddress UnknownHostException]
            [java.security KeyStore]))
-
-(def Broker
-  {:broker-name         (s/maybe s/Str)
-   :authorization-check IFn
-   :inventory           ConcurrentHashMap ;; Mapping of Uri to Connection
-   :metrics-registry    Object
-   :metrics             {s/Keyword Object}
-   :state               Atom})
-
-(s/defn build-and-register-metrics :- {s/Keyword Object}
-  [broker :- Broker]
-  (let [registry (:metrics-registry broker)]
-    (gauges/gauge-fn registry ["puppetlabs.pcp.connections"]
-                     (fn [] (count (:inventory broker))))
-    {:on-connect       (.timer registry "puppetlabs.pcp.on-connect")
-     :on-close         (.timer registry "puppetlabs.pcp.on-close")
-     :on-message       (.timer registry "puppetlabs.pcp.on-message")
-     :on-send          (.timer registry "puppetlabs.pcp.on-send")}))
 
 (defn get-certificate-chain
   "Return the first non-empty certificate chain encountered while scanning the
@@ -84,97 +67,28 @@
 
 ;; connection map lifecycle
 (s/defn add-connection! :- Connection
-  "Add a Connection to the 'inventory' to track a websocket"
+  "Add a Connection to the 'routing-map' to track the websocket and register
+  the change to update 'inventory' when appropriate"
   [broker :- Broker
    connection :- Connection]
   (let [uri (ws->uri (:websocket connection))]
-    (.put (:inventory broker) uri connection)
+    (-> (shared/get-routing-map broker)
+        (.put uri connection))
+    (inventory/add-client broker uri)
     connection))
 
 (s/defn remove-connection!
-  "Remove tracking of a Connection from the broker by websocket"
+  "Remove a Connection from the 'routing-map' and register the change to
+  update 'inventory' when appropriate"
   [broker :- Broker
    uri :- p/Uri]
-  (.remove (:inventory broker) uri))
-
-(s/defn get-connection :- (s/maybe Connection)
-  [broker :- Broker
-   uri :- p/Uri]
-  (get (:inventory broker) uri))
+  (-> (shared/get-routing-map broker)
+      (.remove uri))
+  (inventory/remove-client broker uri))
 
 ;;
 ;; Message processing
 ;;
-
-(def MessageLog
-  "Schema for a loggable summary of a message"
-  {:messageid p/MessageId
-   :source s/Str
-   :messagetype s/Str
-   :destination p/Uri})
-
-(s/defn summarize :- MessageLog
-  [message :- Message]
-  {:messageid (:id message)
-   :messagetype (:message_type message)
-   :source (:sender message)
-   :destination (:target message)})
-
-;; message lifecycle
-
-(s/defn send-message
-  [connection :- Connection
-   message :- Message]
-  (sl/maplog :trace {:type :outgoing-message-trace
-                     :uri (ws->uri (:websocket connection))
-                     :rawmsg message}
-             (i18n/trs "Sending PCP message to '{uri}': '{rawmsg}'"))
-  (websockets-client/send! (:websocket connection)
-                           ((get-in connection [:codec :encode]) message))
-  nil)
-
-(s/defn send-error-message
-  [in-reply-to-message :- (s/maybe Message)
-   description :- s/Str
-   connection :- Connection]
-  (let [error-msg (cond-> (message/make-message
-                           {:message_type "http://puppetlabs.com/error_message"
-                            :sender "pcp:///server"
-                            :data description})
-                    in-reply-to-message (assoc :in_reply_to (:id in-reply-to-message)))]
-    (send-message connection error-msg)))
-
-(s/defn handle-delivery-failure
-  "Send an error message with the specified description."
-  [message :- Message sender :- (s/maybe Connection) reason :- s/Str]
-  (sl/maplog :trace (assoc (summarize message)
-                           :type :message-delivery-failure
-                           :reason reason)
-             (i18n/trs "Failed to deliver '{messageid}' for '{destination}': '{reason}'"))
-  (send-error-message message reason sender))
-
-(s/defn deliver-message
-  "Message consumer. Delivers a message to the websocket indicated by the :target field"
-  [broker :- Broker
-   message :- Message
-   sender :- (s/maybe Connection)]
-  (assert (not (multicast-message? message)))
-  (if-let [connection (get-connection broker (:target message))]
-    (try
-      (sl/maplog
-       :debug (merge (summarize message)
-                     (connection/summarize connection)
-                     {:type :message-delivery})
-       (i18n/trs "Delivering '{messageid}' to '{destination}' at '{remoteaddress}'"))
-      (locking (:websocket connection)
-        (time! (:on-send (:metrics broker))
-               (send-message connection message)))
-      (catch Exception e
-        (sl/maplog :error e
-                   {:type :message-delivery-error}
-                   (i18n/trs "Error in deliver-message"))
-        (handle-delivery-failure message sender (str e))))
-    (handle-delivery-failure message sender (i18n/trs "not connected"))))
 
 (s/defn session-association-request? :- s/Bool
   "Return true if message is a session association message"
@@ -186,7 +100,7 @@
 (s/defn reason-to-deny-association :- (s/maybe s/Str)
   "Returns an error message describing why the session should not be
   allowed, if it should be denied"
-  [broker :- Broker connection :- Connection as :- p/Uri]
+  [_ :- Broker connection :- Connection as :- p/Uri]
   (let [[_ type] (p/explode-uri as)]
     (when (not= type (ws->client-type (:websocket connection)))
       (let [{:keys [uri]} connection]
@@ -226,7 +140,7 @@
    (let [requester-uri (:sender message)
          reason-to-deny (reason-to-deny-association broker connection requester-uri)]
      (process-associate-request! broker message connection reason-to-deny)))
-  ([broker :- Broker
+  ([_ :- Broker
     request :- Message
     connection :- Connection
     reason-to-deny :- (s/maybe s/Str)]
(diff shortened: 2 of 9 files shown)


src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,257 @@
...
+      (loop []
+        (send-updates broker)
+        (if (identical? (deref should-stop 1000 timeout) timeout)
+          (recur))))))
+
+;; N.B. locking order:
+;;   subscriptions
+;;   inventory
+;;   updates
+(s/defn stop-invenotry-updates!
```
wkalt: s/invenotry/inventory here and on 236

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,257 @@
...
+  (let [changes (get-changes broker)
+        inventory (get-inventory broker)
+        updates (get-updates broker)]
+    (locking inventory
+      (locking updates
+        (dotimes [_ (.size changes)]
+          (let [change (.remove changes)]
+            (if (pos? (:change change))
+              (.add inventory (:client change))
+              (.remove inventory (:client change)))
```
wkalt: did we intend to batch updates to the inventory as part of this? I can see that it's set up to return current data to requests, but it seems like we lose something by not being able to use the (:inventory broker) as an accurate representation of the world in code.

I'm a little confused about the necessity for inventory, changes, and updates. I had been imagining we'd have inventory (concurrenthashmap) and changes (linkedlist of [uri change]). add-connection would update both and remove-connection would remove a connection from inventory and enqueue the corresponding change. The scheduled job would then drain the queue, negotiate cancellations and send the resulting list in the update response. What are we getting from the third datastructure?

If we could put both the inventory and the changes queue in a single clojure atom (meaning the concurrenthashmap becomes just a map), we can also get rid of all the explicit locking.

mruzicka: I wanted to avoid locking the concurrenthashmap globally which would need to be done just to get a consistent view of it. So I introduced the `:changes` queue, which is also a concurrent structure (ConcurrentLinkedDeque) and is never locked globally either.
Its purpose is basically to serialize the client connection / disconnections (inventory updates) events. These serialized updates are then pumped into the `:inventory` and `:updates` structures which are OK to lock globally.
These structures are then used to obtain consistent views of the inventory and the update queue, which is needed to ensure updates are not missed / duplicated.

I'm going write up longer PR & commit messages explaining the mechanics in more detail.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: Oops. I've checked all comments but not identifiers.

MikaelSmith:
Looks like we'll need a release of clj-pcp-common.

project.clj
```diff
@@ -38,7 +38,7 @@
                  ;; try+/throw+
                  [slingshot]
 
-                 [puppetlabs/pcp-common "1.0.0"]
+                 [puppetlabs/pcp-common "1.0.1"]
```
MikaelSmith: I'd probably call those changes to pcp-common a `1.1.0` release.

MikaelSmith:
I'll have to spend some time with this, probably Tuesday or Wednesday.

src/puppetlabs/pcp/broker/inventory.clj
wkalt: alright, sorry to jump the gun. I'll comment when the time comes.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -1,6 +1,7 @@
 (ns puppetlabs.pcp.broker.core
-  (:require [metrics.gauges :as gauges]
-            [puppetlabs.experimental.websockets.client :as websockets-client]
+  (:require [puppetlabs.experimental.websockets.client :as websockets-client]
+            [puppetlabs.pcp.broker.shared :as shared :refer
```
MikaelSmith: `shared` isn't a very descriptive name for this.

MikaelSmith: And if you're going to move a bunch of stuff to a new namespace, doing it in a separate commit would help make it easier to review.

src/puppetlabs/pcp/broker/shared.clj
```diff
@@ -0,0 +1,156 @@
...
+        (time! (:on-send (:metrics broker))
+               (send-message connection message)))
+      (catch Exception e
+        (sl/maplog :error e
+                   {:type :message-delivery-error}
+                   (i18n/trs "Error in deliver-message"))
+        (handle-delivery-failure message sender (str e))))
+    (handle-delivery-failure message sender (i18n/trs "not connected"))))
+
+(s/defn deliver-server-message
```
MikaelSmith: This looks like it's handling a different failure: if you try to send an error message to a connection that's closed, you'll get an exception.

MikaelSmith:
Sending a message with `"subscribe": false` doesn't seem to stop inventory updates if you're getting them.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,255 @@
...
+                         (shift-subscription-next-update subscription (-> data meta :next-update-shift)) ;; this returns nil
+                         ;; since we failed to send the update to the client we can't remove the update entries
+                         ;; we failed to send from the update queue so that we can re-try during the next iteration;
+                         ;; so we only allow removal of the entries before the :next-update value of this subscription
+                         ;; (actually the lesser of that value and a value accumulated from previous iterations in
+                         ;; case there were other similarly affected subscriptions)
+                         (min processed-updates-count next-update)))))
+                 processed-updates-count))
+           (count updates)
+           (get-subsribers broker))
```
MikaelSmith: Typo here.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,255 @@
...
+              (HashMap. {:connection   connection
+                         :pattern-sets pattern-sets
+                         :next-update  (-> inventory meta :next-update)}))
+        inventory))))
+
+;; N.B. locking order:
+;;   subscriptions
+;;   inventory
+;;   updates
+(s/defn ^:private get-subsribers :- [p/Uri]
```
MikaelSmith: Should be `get-subscribers`.

src/puppetlabs/pcp/broker/shared.clj
```diff
@@ -0,0 +1,156 @@
...
+           [java.util.concurrent ConcurrentHashMap ConcurrentLinkedDeque]))
+
+(def BrokerState
+  (s/enum :starting :running :stopping))
+
+(def Broker
+  {:broker-name         (s/maybe s/Str)
+   :authorization-check IFn
+   :routing-map         ConcurrentHashMap                   ;; Mapping of Uri to Connection
+   :inventory           HashSet                             ;; Set of known clients (Uris) used for inventory reports
```
MikaelSmith: We'll want to check memory use again, but I get the idea here. We don't want to ever lock `routing-map` or `changes`, but you want to have something that contains a snapshot of the inventory when we send a response. We'll want to re-do some of the memory measurements. My hope is that the set is pretty optimized and just has pointers to the strings also used in routing-map, so it shouldn't add much overhead.

src/puppetlabs/pcp/broker/shared.clj
```diff
@@ -0,0 +1,156 @@
...
+   :changes             ConcurrentLinkedDeque               ;; Queue of pending updates of the :inventory set
+   :updates             LinkedList                          ;; Queue of updates to be sent to the clients subscribed to invetory updates
+   :subscriptions       HashMap                             ;; Mapping of subscribed client Uri to subscription data
```
MikaelSmith: Why use a Java HashMap rather than a Clojure map?

MikaelSmith: Also, I think subscription data should be a `defrecord`.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,255 @@
...
+
+(def PatternSets
+  {:explicit #{p/Uri} :wildcard #{p/ExplodedUri}})
+
+(s/defn init :- Broker
+  [incomplete-broker]                                       ;; the passed broker isn't fully initialized yet
+  (assoc incomplete-broker
+    :inventory     (HashSet.)                               ;; Set of known clients (Uris) used for inventory reports
+    :changes       (ConcurrentLinkedDeque.)                 ;; Queue of pending updates of the :inventory set
+    :updates       (LinkedList.)                            ;; Queue of updates to be sent to the clients subscribed to inventory updates
```
MikaelSmith: Why Java LinkedList instead of a Clojure list? Or vector?

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,255 @@
...
+           [java.util HashSet HashMap LinkedList]
+           [java.util.concurrent ConcurrentLinkedDeque]))
+
+(def PatternSets
+  {:explicit #{p/Uri} :wildcard #{p/ExplodedUri}})
+
+(s/defn init :- Broker
+  [incomplete-broker]                                       ;; the passed broker isn't fully initialized yet
+  (assoc incomplete-broker
+    :inventory     (HashSet.)                               ;; Set of known clients (Uris) used for inventory reports
```
MikaelSmith: Why HashSet instead of Clojure set?

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,255 @@
...
+  [updates :- [p/InventoryChange] subscription :- HashMap]
+  (let [updates-count (count updates)
+        next-update (:next-update subscription)]
+    (if (< next-update updates-count)                       ;; do we have any new updates for this subscription?
+      (let [pattern-sets (:pattern-sets subscription)
+            filtered (->> (subvec updates next-update)      ;; skip updates which have already been sent to this subscription
+                          (reduce
+                            (fn [filtered-updates update]
+                              (if (uri-pattern-sets-match? pattern-sets (:client update))
+                                ;; FIXME there should be a limit on the maximum number of elements
```
MikaelSmith: Why?

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,255 @@
...
+                                ;; in the filtered vector to ensure the update message can't grow
+                                ;; too long
+                                (conj! filtered-updates update)
+                                filtered-updates))
+                            (transient []))
+                          persistent!)
+            next-update-shift (- updates-count next-update)]
+        (if (seq filtered)
+          (with-meta {:changes filtered} {:next-update-shift next-update-shift})
```
MikaelSmith: Did you intend to unwrap this with `meta` somewhere? Or just for testing?

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -370,6 +370,100 @@
                    (is (= "http://puppetlabs.com/inventory_response" (:message_type response)))
                    (is (= [] (:uris data))))))))
 
+(deftest inventory-node-recieves-updates-when-inventory-changes-when-subscribed-to-updates
```
MikaelSmith: I think we could use more test-cases around: subscribing then unsubscribing, sending an inventory update while a node is connecting or disconnecting (we might be able to force a particular sequence using some redefs).

MikaelSmith:
Did some perf testing using https://github.com/puppetlabs/gatling-pcp. Looks solid, very little impact from this change and inventory changes look correct and fast. Tested with 4000 clients and a single inventory subscriber.

MikaelSmith:
@wkalt suggested using something like https://gist.github.com/anonymous/c67214a299ad8632c7f516b7471800ea for the inventory implementation to avoid all the locking. Perf testing in that gist suggests it's worth trying out an implementation.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: This is to prevent us from creating an update message which is longer than the maximum allowed message size, if that happens, the current implementation will die trying to send longer and longer messages because it will never shrink the `:updates` queue.

src/puppetlabs/pcp/broker/shared.clj
mruzicka: I used mutable java data structures to avoid having to wrap them in atoms, which seemed pointless since we do our own concurrent access synchronization/locking.

src/puppetlabs/pcp/broker/shared.clj
mruzicka: Yes, that's exactly the effort here. The snapshots are needed to be able to guarantee we don't miss or duplicate any updates for any subscribed client.

src/puppetlabs/pcp/broker/core.clj
mruzicka: I wasn't planing on this, it was necessitated by a dependency loop I created but did not notice and by the time I did notice it, it was too late to split the commits in any meaningful way.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: it is actually used to compute the number of entries which can be removed from the :updates queue on each iteration

src/puppetlabs/pcp/broker/inventory.clj
MikaelSmith: We don't really know the maximum allowed message size. We could say it's 64MB, and require clients to use that as a minimum.

mruzicka:
@MikaelSmith @wkalt I'm not in principle opposed to using the more idiomatic clojure approach. My concern is that by doing so we might be making things worse during the times, things are already bad - i.e. during mass reconnections of large numbers of clients, like for example after a broker restart.
Could we get a statistics of how many times is the add-connection!/remove-connection! is called without its result being used (because the swap! had to restart)?

mruzicka:
Closing in favor of more idiomatic clojure implementation - https://github.com/puppetlabs/pcp-broker/pull/163

PR #162, opened 2017-01-24
(maint) Fix test client

MikaelSmith:
The test client can't determine when the broker registers a connection
without session association. Force it on tests that depend on multiple
clients being connected at once.


MikaelSmith:
Heh, running with logging enabled exposes some schema violations.

PR #163, opened 2017-02-06
(PCP-600) Inventory updates

mruzicka:
This PR introduces support for inventory updates. To be able to support the updates the `:inventory` map has been replaced with a more complex `:database` map which includes:
- `:inventory`, the original inventory mapping of a client Uri to a websocket connection
- `:updates`, a vector of updates (a vector of inventory change records)
- `:first-update-index`, the index of the first update in the `:updates` vector, this is an ever increasing (and overflowing index) which is increased when entries are removed from the beginning of the `:updates` vector
- `:subscriptions`, mapping of an Uri of a client which subscribed to inventory updates to a subscription record

Each subscription record consists of:
- `:connection`, the websocket connection on which the subscription was received, this is used to detect stale subscriptions
- `:pattern-sets`, a form of Uri filters of the inventory updates the subscriber is interested in receiving
- `:next-update-index`, the index of the next update that should be sent to the subscriber (provided that it matches the :pattern-sets)

The operation:
- when a client connects/disconnects a mapping is added to/removed from the `:inventory` map in the `:database` (if the client is subscribed to the inventory updates on disconnect, it is unsubscribed - removed from the `:subscriptions` map) and an appropriate inventory change record is added to the `:updates` vector
- when an inventory request is received a snapshot of the broker database is obtained while subscribing or unsubscribing the client to inventory updates (adding the client to/removing the the client from the `:subscriptions` map) depending on the `:subscribe` flag in the inventory request, then the `:inventory` from the database snapshot is used to build the inventory response
- updates are sent periodically to subscribed clients; on each iteration a snapshot of the broker database is obtained, then for each subscribed client as per the :subscriptions map in the broker database snapshot the following is performed in a `swap!` call (which purpose is to update the subscription in the broker database):
  - first a check is made whether the subscription from the snapshot is identical to the current subscription, if it's not then the subscription is ignored in this iteration
  - then a check is made whether the update has already been sent (there is an atom outside of the swap! call tracking this information), if it has not then the update message is built and a delivery of the message is attempted; then the atom is updated to indicate the message has been sent
  - finally a new version of the broker database is constructed where the subscription's `:next-update-index` is updated accordingly; this version of the database is returned to the `swap!` call; if the swap doesn't succeed the process is repeated
- after updates are sent to all subscribed clients, processed inventory change records (records which have successfully been sent in inventory updates to all subscribed clients) are removed from the broker database's `:updates` vector and the `:first-update-index` is increased accordingly

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.pcp.broker.core
-  (:require [metrics.gauges :as gauges]
-            [puppetlabs.experimental.websockets.client :as websockets-client]
+  (:require [puppetlabs.experimental.websockets.client :as websockets-client]
+            [puppetlabs.pcp.broker.shared :as shared :refer
+             [Broker get-connection summarize send-error-message send-message deliver-message deliver-server-message]]
             [puppetlabs.pcp.broker.connection :as connection :refer [Codec]]
             [puppetlabs.pcp.broker.websocket :refer [Websocket ws->uri ws->client-type]]
             [puppetlabs.pcp.broker.metrics :as metrics]
@@ -18,28 +19,9 @@
             [slingshot.slingshot :refer [throw+ try+]]
             [puppetlabs.i18n.core :as i18n])
   (:import [puppetlabs.pcp.broker.connection Connection]
-           [clojure.lang IFn Atom]
-           [java.util.concurrent ConcurrentHashMap]
+           [clojure.lang IFn]
            [java.net InetAddress UnknownHostException]
            [java.security KeyStore]))
-
-(def Broker
-  {:broker-name         (s/maybe s/Str)
-   :authorization-check IFn
-   :inventory           ConcurrentHashMap ;; Mapping of Uri to Connection
-   :metrics-registry    Object
-   :metrics             {s/Keyword Object}
-   :state               Atom})
-
-(s/defn build-and-register-metrics :- {s/Keyword Object}
-  [broker :- Broker]
-  (let [registry (:metrics-registry broker)]
-    (gauges/gauge-fn registry ["puppetlabs.pcp.connections"]
-                     (fn [] (count (:inventory broker))))
-    {:on-connect       (.timer registry "puppetlabs.pcp.on-connect")
-     :on-close         (.timer registry "puppetlabs.pcp.on-close")
-     :on-message       (.timer registry "puppetlabs.pcp.on-message")
-     :on-send          (.timer registry "puppetlabs.pcp.on-send")}))
 
 (defn get-certificate-chain
   "Return the first non-empty certificate chain encountered while scanning the
@@ -81,104 +63,34 @@
       (let [message (.getMessage e)]
         (subs message 0 (.indexOf message (int \:)))))))
 
-;; connection map lifecycle
-(s/defn add-connection! :- Connection
-  "Add a Connection to the 'inventory' to track a websocket"
+;; broker database lifecycle
+(s/defn add-connection! :- shared/BrokerDatabase
+  "Add a Connection to the ':inventory' to track the websocket and add
+  a corresponding change record the ':updates' vector."
   [broker :- Broker
    connection :- Connection]
-  (let [uri (ws->uri (:websocket connection))]
-    (.put (:inventory broker) uri connection)
-    connection))
+  (let [database (:database broker)
+        uri (ws->uri (:websocket connection))
+        change {:client uri :change 1}]
+    (swap! database #(-> %
+                         (update :inventory assoc uri connection)
+                         (update :updates conj change)))))
 
-(s/defn remove-connection!
-  "Remove tracking of a Connection from the broker by websocket"
+(s/defn remove-connection! :- shared/BrokerDatabase
+  "Remove a Connection from ':inventory' and possibly ':subscriptions' and
+  add a corresponding change record the ':updates' vector."
   [broker :- Broker
    uri :- p/Uri]
-  (.remove (:inventory broker) uri))
-
-(s/defn get-connection :- (s/maybe Connection)
-  [broker :- Broker
-   uri :- p/Uri]
-  (get (:inventory broker) uri))
+  (let [database (:database broker)
+        change {:client uri :change -1}]
+    (swap! database #(-> %
+                         (update :inventory dissoc uri)
+                         (update :subscriptions dissoc uri)
+                         (update :updates conj change)))))
 
 ;;
 ;; Message processing
 ;;
-
-(def MessageLog
-  "Schema for a loggable summary of a message"
-  {:messageid p/MessageId
-   :source s/Str
-   :messagetype s/Str
-   :destination p/Uri})
-
-(s/defn summarize :- MessageLog
-  [message :- Message]
-  {:messageid (:id message)
-   :messagetype (:message_type message)
-   :source (:sender message)
-   :destination (:target message)})
-
-;; message lifecycle
-
-(s/defn send-message
-  [connection :- Connection
-   message :- Message]
-  (sl/maplog :trace {:type :outgoing-message-trace
-                     :uri (ws->uri (:websocket connection))
-                     :rawmsg message}
-             (i18n/trs "Sending PCP message to '{uri}': '{rawmsg}'"))
-  (websockets-client/send! (:websocket connection)
-                           ((get-in connection [:codec :encode]) message))
-  nil)
-
-(s/defn send-error-message
-  [in-reply-to-message :- (s/maybe Message)
-   description :- s/Str
-   connection :- Connection]
-  (let [error-msg (cond-> (message/make-message
-                           {:message_type "http://puppetlabs.com/error_message"
-                            :sender "pcp:///server"
-                            :data description})
-                    in-reply-to-message (assoc :in_reply_to (:id in-reply-to-message)))]
-    (try
-      (send-message connection error-msg)
-      (catch Exception e
-        (sl/maplog :debug e
-                   {:type :message-delivery-error}
-                   (i18n/trs "Error in send-error-message"))))))
-
-(s/defn handle-delivery-failure
-  "Send an error message with the specified description."
-  [message :- Message sender :- (s/maybe Connection) reason :- s/Str]
-  (sl/maplog :trace (assoc (summarize message)
-                           :type :message-delivery-failure
-                           :reason reason)
-             (i18n/trs "Failed to deliver '{messageid}' for '{destination}': '{reason}'"))
-  (send-error-message message reason sender))
-
-(s/defn deliver-message
-  "Message consumer. Delivers a message to the websocket indicated by the :target field"
-  [broker :- Broker
-   message :- Message
-   sender :- (s/maybe Connection)]
-  (assert (not (multicast-message? message)))
-  (if-let [connection (get-connection broker (:target message))]
-    (try
-      (sl/maplog
-       :debug (merge (summarize message)
-                     (connection/summarize connection)
-                     {:type :message-delivery})
-       (i18n/trs "Delivering '{messageid}' to '{destination}' at '{remoteaddress}'"))
-      (locking (:websocket connection)
-        (time! (:on-send (:metrics broker))
-               (send-message connection message)))
-      (catch Exception e
-        (sl/maplog :error e
-                   {:type :message-delivery-error}
-                   (i18n/trs "Error in deliver-message"))
-        (handle-delivery-failure message sender (str e))))
-    (handle-delivery-failure message sender (i18n/trs "not connected"))))
 
 (s/defn session-association-request? :- s/Bool
   "Return true if message is a session association message"
@@ -190,7 +102,7 @@
 (s/defn reason-to-deny-association :- (s/maybe s/Str)
   "Returns an error message describing why the session should not be
   allowed, if it should be denied"
-  [broker :- Broker connection :- Connection as :- p/Uri]
+  [_ :- Broker connection :- Connection as :- p/Uri]
   (let [[_ type] (p/explode-uri as)]
     (when (not= type (ws->client-type (:websocket connection)))
       (let [{:keys [uri]} connection]
@@ -230,7 +142,7 @@
    (let [requester-uri (:sender message)
          reason-to-deny (reason-to-deny-association broker connection requester-uri)]
      (process-associate-request! broker message connection reason-to-deny)))
-  ([broker :- Broker
+  ([_ :- Broker
     request :- Message
     connection :- Connection
     reason-to-deny :- (s/maybe s/Str)]
@@ -261,11 +173,6 @@
          nil)
        (assoc connection :uri requester-uri)))))
 
-(s/defn make-inventory_response-data-content :- p/InventoryResponse
-  [broker query]
-  {:uris (doall (filter (partial get-connection broker)
-                        (inventory/find-clients broker query)))})
-
 (s/defn process-inventory-request
   "Process a request for inventory data.
    This function assumes that the requester client is associated.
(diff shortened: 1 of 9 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -81,105 +63,35 @@
...
+  (let [database (:database broker)
+        uri (ws->uri (:websocket connection))
+        change {:client uri :change 1}]
+    (swap! database #(-> %
+                         (update :inventory assoc uri connection)
+                         (update :updates conj change)))))
+
+(s/defn remove-connection! :- shared/BrokerDatabase
+  "Remove a Connection from ':inventory' and possibly ':subscriptions' and
+  remember the change tsend it to subscribed clients when appropriate"
```
MikaelSmith: Broken sentence.

MikaelSmith:
Initial perf test looks pretty good:
```
================================================================================
---- Global Information --------------------------------------------------------
> request count                                      27000 (OK=27000  KO=0     )
> min response time                                      2 (OK=2      KO=-     )
> max response time                                   2861 (OK=2861   KO=-     )
> mean response time                                    44 (OK=44     KO=-     )
> std deviation                                        209 (OK=209    KO=-     )
> response time 50th percentile                          4 (OK=4      KO=-     )
> response time 75th percentile                          5 (OK=5      KO=-     )
> response time 95th percentile                         47 (OK=47     KO=-     )
> response time 99th percentile                       1038 (OK=1038   KO=-     )
> mean requests/sec                                    150 (OK=150    KO=-     )
---- Response Time Distribution ------------------------------------------------
> t < 800 ms                                         26226 ( 97%)
> 800 ms < t < 1200 ms                                 575 (  2%)
> t > 1200 ms                                          199 (  1%)
> failed                                                 0 (  0%)
================================================================================
```

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+  at the time of subscribing."
+  [broker :- Broker client :- p/Uri connection :- Connection pattern-sets :- shared/PatternSets]
+  (let [database (:database broker)]
+    (swap! database
+           #(update % :subscriptions assoc client {:connection        connection
+                                                   :pattern-sets      pattern-sets
+                                                   :next-update-index (unchecked+ (-> % :first-update-index) (-> % :updates count))}))))
+
+(s/defn unsubscribe-client! :- shared/BrokerDatabase
+  "Unsubscribe the specified client form inventory updates. Return the broker database snapshot
```
MikaelSmith: Typo.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+                                         (if (> processed-count next-update-offset)
+                                           ;; add the count of the InventoryChange records which were processed when building the
+                                           ;; inventory update for this subscriber to the :next-update-index in the subscription
+                                           (update database :subscriptions assoc subscriber
+                                                   (update subscription :next-update-index unchecked+ (- processed-count next-update-offset)))
+                                           database)))))) ;; no need to update the subscription if nothing past the next-update-offset was processed
+                 (min @processed-count-atom processed-updates-count))
+               processed-updates-count))
+           updates-count
+           (-> database-snapshot :subscriptions keys))
```
MikaelSmith: You could iterate over `(:subscriptions database-snapshot)`, which would give you `subscriber` and `subscription` at the same time (no need to immediately do a lookup). They'd be given as a vector you can destructure.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+  "Send inventory update messages to subscribed clients. Subsequently remove the processed inventory
+  change records from the updates queue."
+  (let [database (:database broker)
+        database-snapshot (deref database)
+        updates (:updates database-snapshot)
+        updates-count (count updates)
+        first-update-index (:first-update-index database-snapshot)]
+    (->> (reduce                                            ;; this reduce returns the number of updates which can be removed from the updates queue
+           (fn [processed-updates-count subscriber]
+             (if-let [subscription (-> broker :database deref :subscriptions (get subscriber))]
```
MikaelSmith: Could be `(-> database deref :subscriptions (get subscriber)`, but that's a nitpic.

src/puppetlabs/pcp/broker/inventory.clj
MikaelSmith: Ok, now I see that subscriptions is from the snapshot, but you check the live database for each subscriber to ensure it's still subscribed. That makes sense.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: 👍

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+                                           database))) ;; no need to update the subscription if nothing past the next-update-offset was processed
+                                     (do
+                                       ;; if the subscription has been changed, ignore it but pretend the entire updates vector was processed
+                                       (reset! processed-count-atom updates-count)
+                                       database))))
+                 (min @processed-count-atom processed-updates-count))
+               processed-updates-count))
+           updates-count
+           (-> database-snapshot :subscriptions keys))
+         (remove-processed-updates broker))))
```
wkalt: it would be good to break this function up if possible

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+
+(s/defn start-inventory-updates!
+  "Start periodic sending of the inventory updates."
+  [broker :- Broker]
+  (future
+    (let [should-stop (:should-stop broker)
+          timeout (Object.)]
+      (loop []
+        (send-updates broker)
+        (if (identical? (deref should-stop 1000 timeout) timeout)
```
wkalt: The objects seem unnecessary I think -- couldn't you just have
```
(if (nil? (deref should-stop 1000 nil))
...
)
```
?

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+          (recur))))))
+
+(s/defn stop-inventory-updates!
+  "Stop the periodic sending of the inventory updates."
+  [broker :- Broker]
+  (let [database (:database broker)]
+    (swap! database assoc :subscriptions {}))         ;; clear all subscriptions
+  (deliver (:should-stop broker) nil))
```
wkalt: ah, I see. Maybe we should deliver true instead of nil?

src/puppetlabs/pcp/broker/shared.clj
```diff
@@ -0,0 +1,171 @@
...
+   sender :- (s/maybe Connection)]
+  (assert (not (multicast-message? message)))
+  (if-let [connection (get-connection broker (:target message))]
+    (try
+      (sl/maplog
+        :debug (merge (summarize message)
+                      (connection/summarize connection)
+                      {:type :message-delivery})
+        (i18n/trs "Delivering '{messageid}' to '{destination}' at '{remoteaddress}'"))
+      (locking (:websocket connection)
```
wkalt: we're inconsistent in whether we lock the websocket connection or use try/catch to guard against issues with this, e.g
https://github.com/puppetlabs/pcp-broker/pull/163/files#diff-ea6951d0c7549e5dffcdc11b13bfaca0R102

Is there a reason for that? My impulse would be to stick with try/catch if possible.

src/puppetlabs/pcp/broker/shared.clj
```diff
@@ -0,0 +1,171 @@
...
+(s/defn deliver-server-message
+  "Message consumer. Delivers a message to the websocket indicated by the :target field but only if it still
+  routed to the conection specified by the client argument"
+  [broker :- Broker
+   message :- Message
+   client :- Connection]
+  (assert (not (multicast-message? message)))
+  (let [connection (get-connection broker (:target message))
+        message (assoc message :sender "pcp:///server")]
+    (if (identical? connection client)
```
wkalt: why not just use client?

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+                                         (let [data (-> (subvec updates next-update-offset) ;; skip updates which have already been sent to this subscriber
+                                                        (build-update-data (:pattern-sets subscription)))]
+                                           (if (or (nil? data) ;; there are no updates for this subscriber
+                                                   (deliver-server-message broker
+                                                                           (message/make-message
+                                                                             {:message_type "http://puppetlabs.com/inventory_update"
+                                                                              :target subscriber
+                                                                              :data data})
+                                                                           (:connection subscription)))
+                                             (reset! processed-count-atom updates-count)
```
wkalt: I think updates-count may be wrong, as it could have changed between lines 122 and 129

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+                     (reduce
+                       (fn [matched-clients [client _]]
+                         (if (uri-pattern-sets-match? pattern-sets client)
+                           (conj! matched-clients client)
+                           matched-clients))
+                       (transient #{}))
+                     persistent!)]
+    {:uris (sort matched)}))
+
+(s/defn build-update-data :- (s/maybe p/InventoryUpdate)
```
wkalt: there was some discussion at some point around this canceling out reconnections -- would be good to capture that in another ticket if we're not doing it here

src/puppetlabs/pcp/broker/inventory.clj
MikaelSmith: I don't think that's true, updates-count is based on a snapshot so the number of updates shouldn't change.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -1,20 +1,177 @@
...
+               (let [next-update-offset (unchecked- (:next-update-index subscription) first-update-index)
+                     processed-count-atom (atom nil)]       ;; how many records from the updates vector were processed
+                 (swap! database (fn [database]
+                                   (if (identical? (-> database :subscriptions (get subscriber)) subscription)
+                                     (do
+                                       (if (nil? @processed-count-atom) ;; have we not sent the update to this subscriber yet?
+                                         (let [data (-> (subvec updates next-update-offset) ;; skip updates which have already been sent to this subscriber
+                                                        (build-update-data (:pattern-sets subscription)))]
+                                           (if (or (nil? data) ;; there are no updates for this subscriber
+                                                   (deliver-server-message broker
```
wkalt: we should do any sending/side-effecting stuff outside of the swap!, since the swap! function body will often get executed multiple times.

wkalt: see https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Atom.java#L33 for reference

wkalt: not that this also applies to the modifications to processed-count-atom in here -- those changes will not be rolled back if the swap on line 129 is rolled back

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: Yeah, that was just me being paranoid, to ensure you can't set the promise to the value which is returned when the timeout expires and thus cause an infinite loop - delivering `nil` to the promise in your example would have this effect.
But I admit it's probably too paranoid.

src/puppetlabs/pcp/broker/shared.clj
mruzicka: There is a potential that by the time you get around to actually sending the message, the connection which was used for subscribing to the periodic updates was closed and replaced with a new one, and in this situation you don't want to send the inventory update. Well you can try to do that, but it should fail, because the original `client` connection should be closed at that point. Having the test there avoids avoids the attempt which is bound to fail and makes it possible to provide a better diagnostics.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: All of that is intentional, the `processed-count-atom` atom is actually used to guard against the duplicate message send. Notice that the message is only sent when the atom is at its initial value - `nil`. When the message is sent (or there is nothing to send) the `processed-count-atom` atom is reset to a non-nil value, preventing any duplicate send of the update in case the swap! fails and re-executes the update function.
The other purpose of the `processed-count-atom` atom is to get the count of InventoryUpdate records processed for the particular subscription out of the swap! call context.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: Correct, the code only attempts to send updates from the snapshot, not anything which might have been added to the live database after the snapshot was taken. Those entries will be processed in the next iteration.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: I'll see what I can do, but it uses a couple of values from its enclosing scope, so it might need to take a bunch of arguments when moved out of that scope.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: That makes sense.

src/puppetlabs/pcp/broker/shared.clj
mruzicka: An exception thrown in the `devliver-server-message` would terminate the thread sending the updates, so the try/catch is a necessity. I'll add that.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: If I'm counting correctly the function would have to take 5 arguments, which makes not want to change the current code.

src/puppetlabs/pcp/broker/shared.clj
mruzicka: Actually on a closer look, there is already a try / catch in the `devliver-server-message` function - [here](https://github.com/mruzicka/pcp-broker/blob/afca982969cf2b6f620384d931f5eed7d3cf0cfa/src/puppetlabs/pcp/broker/shared.clj#L156)

mruzicka: I believe the `(locking )` is there because of the metrics - see [here](https://github.com/mruzicka/pcp-broker/blob/afca982969cf2b6f620384d931f5eed7d3cf0cfa/src/puppetlabs/pcp/broker/shared.clj#L163).

src/puppetlabs/pcp/broker/inventory.clj
wkalt: thanks, can see that now

src/puppetlabs/pcp/broker/inventory.clj
wkalt: ok -- typically once a function is too wide to view in github I start thinking about breaking it up. Up to you though.

src/puppetlabs/pcp/broker/shared.clj
wkalt: ok -- I guess what I was getting at is it seems like it could all just be handled in the catch (i.e the client disconnected case as well as any other failure). I guess this works though.

wkalt:
@mruzicka left a few minor comments but I think this is ok. Thanks for talking through the correctness concerns I had -- those were not warranted in the end.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: I'v logged [PCP-706](https://tickets.puppetlabs.com/browse/PCP-706) to track this.

src/puppetlabs/pcp/broker/inventory.clj
mruzicka: good rule of thumb, I've at least decreased the indentation of the code by inserting a line break

mruzicka:
@wkalt @MikaelSmith I've addressed the comments, this should be now good to merge.

MikaelSmith:
Created [PCP-707](https://tickets.puppetlabs.com/browse/PCP-707) to expand tests for higher concurrency scenarios.

src/puppetlabs/pcp/broker/shared.clj
wkalt: ah

PR #164, opened 2017-02-10
(PCP-681) Add outgoing connections for controllers

MikaelSmith:
Adds the ability to configure outgoing websocket connections, along with a whitelist to limit what messages they can send.

This PR isn't as big as it looks, most of the lines are regenerating SSL certs. Work through individual commits.


MikaelSmith:
I plan to add one more test for multiple controllers. This seems ready for review though.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -532,6 +610,7 @@
 
 (s/defn stop
   [broker :- Broker]
+  (doseq [[_ client] @(:controllers broker)] (pcp-client/close (:websocket client)))
```
wkalt: this might be a moot point, but should we also stop accepting new connections when a stop is called?

wkalt: oh right, these are outbound anyway

MikaelSmith: We already do that, by setting the state to stopping.

wkalt: if you're referring to line 614, seems like if we're relying on that to prohibit new connections it should be moved above this line right?

MikaelSmith: That seems reasonable.

MikaelSmith:
Ok, I think this is complete.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -490,6 +495,79 @@
...
+(s/defn start-client
+  [broker :- Broker
+   ssl-context :- SSLContext
+   controller-whitelist :- #{s/Str}
+   uri :- s/Str]
+  (let [client (pcp-client/connect {:server uri
+                                    :ssl-context ssl-context
+                                    :type "server"}
+                                    {:default (partial default-message-handler broker controller-whitelist)})
+        uri-obj (URI. uri)
```
MikaelSmith: This places a requirement on the Orchestrator node to use a hostname that conforms to a Java URI. Note that some environments use hostnames with underscores in them, which are not considered valid URI according to Java's implementation. See also https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_hostnames.

mruzicka:
@MikaelSmith I have just a couple of nitpick comments, I'm fine merging as is. Please let me know whether you are going to address any of those comments.

src/puppetlabs/pcp/broker/connection.clj
```diff
@@ -36,16 +34,16 @@
...
+    :uri            uri}))
 
 (s/defn -summarize :- ConnectionLog
   [connection :- Connection]
-  (let [{:keys [common-name remote-address]} connection]
-    {:commonname common-name
-     :remoteaddress remote-address}))
+  (let [{:keys [websocket]} connection]
+    {:commonname    (ws->common-name websocket)
+     :remoteaddress (ws->remote-address websocket)}))
```
MikaelSmith: I can give that a try.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -423,21 +426,25 @@
...
-    (let [connection (connection/make-connection ws codec)
-          {:keys [common-name]} connection]
-      (if (nil? common-name)
+    ;; Use a dummy Uri until we verify a real one can be constructed.
+    (let [connection (connection/make-connection ws codec "pcp:///dummy")]
+      (if (nil? (ws->common-name ws))
         (do
           (sl/maplog :debug (assoc (connection/summarize connection)
                                    :type :connection-no-peer-certificate)
                      (i18n/trs "No client certificate, closing '{remoteaddress}'"))
```
MikaelSmith: Seems reasonable.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -532,6 +592,7 @@
 
 (s/defn stop
   [broker :- Broker]
+  (doseq [[_ client] @(:controllers broker)] (pcp-client/close (:websocket client)))
   (-> broker :state (reset! :stopping))
```
MikaelSmith: I did, maybe that ended up in the wrong commit.

src/puppetlabs/pcp/broker/connection.clj
```diff
@@ -36,16 +34,16 @@
...
   [websocket :- Websocket
-   codec :- Codec]
+   codec :- Codec
+   uri :- p/Uri]
   ;; NOTE(ale): the 'map->...' constructor comes from schema.core's defrecord
   (map->Connection
    {:websocket      websocket
     :codec          codec
-    :remote-address (ws->remote-address websocket)
-    :common-name    (ws->common-name websocket)}))
```
MikaelSmith: That's not feasible for how it's handled with Client. I'll try some measurements to see whether it has a significant impact when debug logging is enabled.

src/puppetlabs/pcp/broker/connection.clj
```diff
@@ -36,16 +34,16 @@
...
+    :uri            uri}))
 
 (s/defn -summarize :- ConnectionLog
   [connection :- Connection]
-  (let [{:keys [common-name remote-address]} connection]
-    {:commonname common-name
-     :remoteaddress remote-address}))
+  (let [{:keys [websocket]} connection]
```
MikaelSmith: When I change this, I start getting null pointer exceptions in the mock server shutdown, but only when schema validation is enabled. Not making sense yet.

MikaelSmith:
Updated everything except the websocket destructuring. Still figuring out why that messes up tests.

src/puppetlabs/pcp/broker/connection.clj
mruzicka: weird

src/puppetlabs/pcp/broker/connection.clj
mruzicka: ok

src/puppetlabs/pcp/broker/connection.clj
MikaelSmith: Figured it out. It's purely a timing issue, not sure why it was so consistent when I change this. If the client closes its side of the connection first, the underlying Jetty session can close and cause websocket-client/close! to throw a null pointer exception.

MikaelSmith:
Should be all fixed up now.

src/puppetlabs/pcp/broker/connection.clj
MikaelSmith: Ran gatling-pcp, I don't see any appreciable change in performance with those tests. The extra cost to remote-address and common-name should be pretty modest: a few extra indirections, and generally only happen once or twice per event.

src/puppetlabs/pcp/broker/connection.clj
mruzicka: OK, do we handle that in the production code (as opposed to the test code)?

MikaelSmith: Yes. Although usually it's jetty catching the exception and ending that websocket callback, so maybe we could be a little friendlier about it.

PR #165, opened 2017-02-14
Inline client

MikaelSmith:


MikaelSmith:
Ignoring this until we have a better reason to do something like this.

PR #167, opened 2017-02-18
(PCP-707, PCP-600) Add stress test for subscriptions, fix race that sends inventory update before response

MikaelSmith:
# (maint) Refactor integration tests

Extract common service setup into a new testutils namespace, and split
subscription tests into their own namespace.

Also removes unused library inclusion and fixes a race in finishing
`it-closes-connections-when-not-running-test`.

# (PCP-700) Add stress test for subscriptions and many connections

Adds a stress test for many connections being established as
subscriptions are initiated. Ensures inventory stays insync.

# (PCP-600) Fix race sending inventory update before requests

Normally a client requesting updates is registered as a subscriber, and
an inventory response sent. It happens in this order to ensure no
updates are missed; the reverse order would allow for an inventory
response to be sent then a change in inventory to happen before the
subscription is registered.

However, the current order allows for an inventory update to be sent by
the subscription thread between registering as a subscriber and sending
the inventory response. Use locking to ensure that subscription and
inventory response happen in sequence, preventing any updates from being
sent to the subscriber until the response has first been sent.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -187,8 +187,9 @@
     (s/validate p/InventoryRequest data)
     (let [requester-uri (:sender message)
           pattern-sets (inventory/build-pattern-sets (:query data))
+          connection-promise (promise)
           database (if (:subscribe data)
-                     (inventory/subscribe-client! broker requester-uri connection pattern-sets)
+                     (inventory/subscribe-client! broker requester-uri connection-promise pattern-sets)
                      (inventory/unsubscribe-client! broker requester-uri))
           data (-> database :inventory (inventory/build-inventory-data pattern-sets))
           response (message/make-message
@@ -196,7 +197,10 @@
                       :target requester-uri
                       :in_reply_to (:id message)
                       :data data})]
-      (deliver-server-message broker response connection)))
+      (try
+        (deliver-server-message broker response connection)
+        (finally
+          (deliver connection-promise connection)))))
   nil)
 
 (s/defn process-server-message! :- (s/maybe Connection)
--- a/src/puppetlabs/pcp/broker/inventory.clj
+++ b/src/puppetlabs/pcp/broker/inventory.clj
@@ -159,6 +159,8 @@
            (-> database-snapshot :subscriptions keys))
          (remove-processed-updates broker))))
 
+(def batch-update-interval-ms 1000)
+
 (s/defn start-inventory-updates!
   "Start periodic sending of the inventory updates."
   [broker :- Broker]
@@ -166,7 +168,7 @@
     (let [should-stop (:should-stop broker)]
       (loop []
         (send-updates broker)
-        (if (nil? (deref should-stop 1000 nil))
+        (if (nil? (deref should-stop batch-update-interval-ms nil))
           (recur))))))
 
 (s/defn stop-inventory-updates!
--- a/src/puppetlabs/pcp/broker/inventory.clj
+++ b/src/puppetlabs/pcp/broker/inventory.clj
@@ -85,9 +85,10 @@
           {:changes filtered})))
 
 (s/defn subscribe-client! :- shared/BrokerDatabase
-  "Subscribe the specified client for inventory updates. Return the broker database snapshot
-  at the time of subscribing."
-  [broker :- Broker client :- p/Uri connection :- Connection pattern-sets :- shared/PatternSets]
+  "Subscribe the specified client for inventory updates. Expects a promise of a connection, to be
+  fulfilled when the initial inventory response has been sent; that prevents messages appearing out
+  of order. Return the broker database snapshot at the time of subscribing."
+  [broker :- Broker client :- p/Uri connection :- Object pattern-sets :- shared/PatternSets]
   (let [database (:database broker)]
     (swap! database
            #(update % :subscriptions assoc client {:connection        connection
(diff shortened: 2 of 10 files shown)


MikaelSmith:
@mruzicka ran into one race building this test. Would appreciate a review of it, and whether you think there are other test scenarios we should build.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -186,17 +186,21 @@
...
-                     {:message_type "http://puppetlabs.com/inventory_response"
-                      :target requester-uri
-                      :in_reply_to (:id message)
-                      :data data})]
-      (deliver-server-message broker response connection)))
+          pattern-sets (inventory/build-pattern-sets (:query data))]
+      ;; Lock the websocket connection. This is done to ensure that no updates are sent to this
+      ;; connection before the inventory response is sent. Without locking, that could happen because
+      ;; we subscribe before sending the response.
+      (locking (:websocket connection)
```
mruzicka: I'd prefer to find other solution than locking, as that can block sending updates to other subscribed clients.
Perhaps not setting the ```:connection``` key in the subscription map until after the inventory response is sent. And test for that eventuality in the code sending the updates.

MikaelSmith: Hmm. As long as we have the correct updates index registered, we could skip sending updates until the connection is available. We definitely need to track any updates that happen after the snapshot. I may have a simple way to address this.

MikaelSmith: Updated.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -87,13 +87,21 @@
...
     (swap! database
-           #(update % :subscriptions assoc client {:connection        connection
+           #(update % :subscriptions assoc client {:connection        nil
                                                    :pattern-sets      pattern-sets
                                                    :next-update-index (unchecked+ (-> % :first-update-index) (-> % :updates count))}))))
 
+(s/defn enable-client! :- shared/BrokerDatabase
+  "Enable sending inventory updates to the specified client. This is done after subscribing to
+  ensure the inventory request is the first message sent back to the client."
+  [broker :- Broker client :- p/Uri connection :- Connection]
```
mruzicka: The function needs to take the subscription map as a parameter. It needs to verify in the `(swap! )` call, that the subscription currently in database (if any) is the same as the one passed in the parameter. Otherwise you might replace the `:connection` entry on a subscription which superseded the original one.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -134,6 +142,7 @@
                                 (let [data (-> (subvec updates next-update-offset) ;; skip updates which have already been sent to this subscriber
                                                (build-update-data (:pattern-sets subscription)))]
                                   (if (or (nil? data) ;; there are no updates for this subscriber
+                                          (nil? (:connection subscription)) ;; prevent update before inventory response
                                           (deliver-server-message broker
                                                                   (message/make-message
```
mruzicka: the condition needs to use an `(and )`:
```clojure
                                  (if (or (nil? data) ;; there are no updates for this subscriber
                                          (and (nil? (:connection subscription))
                                               (deliver-server-message broker
                                                                       (message/make-message
...
```
as we need to re-process the inventory changes not reflected in the inventory response sent to the subscriber upon subscribing - some of the changes might be missed if we allow deleting more entries from the update vector than `next-update-offset`

mruzicka: There is a similar problem with this condition above (not part of the diff):
```clojure
(if (identical? (-> database :subscriptions (get subscriber)) subscription)
```

mruzicka: I wonder if we should use a ```(with-meta )``` to attach a flag to the subscription (telling whether the inventory has already been sent or not) rather than not including the `:connection` in the subscription.

MikaelSmith: Would that avoid changing the object, so we can continue using `identical?`?

mruzicka: But we could not use `(identical? )` then, as `(with-meta )` creates a new object, which does not hold any reference to the original one.

mruzicka: Perhaps the easiest would be to include a `(promise)` in the subscription, which would only be delivered when the inventory response is sent ...

MikaelSmith: Ah, that might work. I'll work on tests to expose this issue too.

mruzicka: With the `(promise)` approach we would not need to modify the broker database on client subscription more that once, and we would not need to modify the logic in the `(send-updates )` beyond checking whether the response has been delivered:
```clojure
...
            (and (realized? (:response-sent subscription))
                 (deliver-server-message broker
                                         (message/make-message
...
```

MikaelSmith: I believe I fixed this, and added two tests to explicitly cover some of the problem areas we identified.

src/puppetlabs/pcp/broker/inventory.clj
```diff
@@ -133,13 +134,14 @@
...
-                                          (deliver-server-message broker
-                                                                  (message/make-message
-                                                                    {:message_type "http://puppetlabs.com/inventory_update"
-                                                                     :target subscriber
-                                                                     :data data})
-                                                                  (:connection subscription)))
+                                  (if (and (realized? (:connection subscription)) ;; prevent update before inventory response
+                                           (or (nil? data) ;; there are no updates for this subscriber
+                                               (deliver-server-message broker
+                                                                       (message/make-message
```
mruzicka: this can be slightly optimized by rearranging the condition to:
```clojure
      (if (or (nil? data) ;; there are no updates for this subscriber
              (and (realized? (:connection subscription)) ;; prevent update before inventory response
                   (deliver-server-message broker
                                           (message/make-message
```
as when there is no updates to send to the subscriber (no updates matched the filter on the subscription), we can safely remove the processed entries from the `:updates` vector even though no update was sent because no update would be sent regardless as there was nothing to send

PR #168, opened 2017-02-23
(maint) Subscription improvements

MikaelSmith:
Some new tests and and a minor improvement around subscriptions.

Also identified a problem with controllers remaining subscribed after they disconnect. Created [PCP-712](https://tickets.puppetlabs.com/browse/PCP-712).


MikaelSmith:
Removed the test expected to show failure to unsubscribe because it was flaky. Fixed up in the PR that fixes that issue.

PR #171, opened 2017-02-28
(maint) Accept 1006 for connection failure due to authz rules

MikaelSmith:
Sometimes the test related to confirming that connections rejected due
to authorization rules receives a 1006 error code. This may be happening
because we close the connection during the on-connect callback.

I don't see a straight-forward way to reject the connection otherwise,
and failure to connect isn't something we should return detailed
information about, so accept this failure mode.


MikaelSmith:
Filed https://tickets.puppetlabs.com/browse/PCP-714 to review this later.

PR #173, opened 2017-03-06
(maint) Remove authorization.conf file to avoid conflicts

nicklewis:
We use inconsistent names for this file across different packages. In order
to include pcp-broker in multiple packages where that filename is
different, we have no good choice but to remove it from this jar. The file
has been moved to test-resources to support existing dev workflows.


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.1.1
+
+This is a maintenance releqase
```
MikaelSmith: Typo

PR #176, opened 2017-03-09
(PCP-701) disallow client connections until a controller is connected

wkalt:


src/puppetlabs/pcp/broker/service.clj
```diff
@@ -48,7 +50,11 @@
                                :ssl-context-factory
                                .getSslContext)
                broker (assoc broker :broker-name broker-name)
-               context (assoc context :broker broker)]
+               context (assoc context :broker broker)
+               timestamp (t/now)
+               controller-pcp-uris (map #(str/replace % #"wss" "pcp") controller-uris)]
+           (swap! (:database broker) assoc :warning-bin
```
MikaelSmith: Why here instead of in initiate-controller-connections?

MikaelSmith:
We should also update the pcp-broker to report a degraded status if all controllers are disconnected as well.

src/puppetlabs/pcp/broker/service.clj
wkalt: it seemed separate from connection initiation to me. I would have liked to stick it in database initialization, but the code isn't well set up to do that right now.

MikaelSmith: Ok, not unreasonable just seems to expose something that was an internal interest.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -660,8 +662,11 @@
 
 (s/defn start
   [broker :- Broker]
-  (inventory/start-inventory-updates! broker)
-  (-> broker :state (reset! :running)))
+  (let [initial-state (if (empty? @(:controllers broker))
```
MikaelSmith: We start the broker after we initiate controller connections, meaning this could be incorrectly left in an `:error` state if the controller connection is established before `start` is called.

MikaelSmith: How about just querying `all-controllers-disconnected?` when `status` is called?

PR #177, opened 2017-03-09
(maint) always run tests with schema validation

wkalt:


MikaelSmith:
You'll want to update the Travis config as well; this changed doubled the time because it's running everything twice.

wkalt:
@MikaelSmith I updated travis.yaml. The test failures seem unrelated but probably warrant some investigation.

MikaelSmith:
Oh, there's a good reason to test without schema validation. Previously we had relied on schemas to catch errors, and our tests started failing when we switched schema validation off by default. So we started testing both.

wkalt:
@MichaelSmith this will test both on travis now

wkalt:
@MikaelSmith rather

MikaelSmith:
@mruzicka do you have any thoughts on this?

mruzicka:
@MikaelSmith @wkalt I'd prefer running the tests with the schema checks on only. The reason we gave up on that was that the tests were occasionally failing due to timeouts.

MikaelSmith:
We run tests in CI with schema checks both enabled and disabled. Each can potentially catch issues.

This would be inconsistent with some other projects, and I don't see a strong use case for it (it's easy to run locally if CI failed).

PR #178, opened 2017-03-10
(PCP-701) Fix state to report starting and stopping

MikaelSmith:
In a previous commit, status was changed to report error if all
controllers were disconnected. However, this is also normal behavior
when starting or stopping the service, which made status less
illuminating. Restore reporting starting and stopping status, and only
report error if the broker is running and all controllers are
disconnected.

### DIFF
--- a/test/integration/puppetlabs/pcp/broker/controllers_test.clj
+++ b/test/integration/puppetlabs/pcp/broker/controllers_test.clj
@@ -328,3 +328,20 @@
                     (testing "returns running state when brokers are disconnected"
                       (= :running (:state (core/status broker :info))))
                     (is (not (empty? (:inventory @database)))))))))))))
+
+(deftest status-with-controller
+  (let [start-fn core/start
+        stop-fn core/stop]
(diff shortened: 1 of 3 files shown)


test/integration/puppetlabs/pcp/broker/controllers_test.clj
```diff
@@ -328,3 +328,20 @@
...
+(deftest status-with-controller
+  (let [start-fn core/start
+        stop-fn core/stop]
+    (with-redefs [core/start (fn [b]
+                               (is (= :starting (:state (core/status b :info))))
+                               (start-fn b))
+                  core/stop (fn [b]
+                              (let [r (stop-fn b)]
+                                (is (= :stopping (:state (core/status b :info))))
+                                r))]
```
wkalt: why return r here?

MikaelSmith: Well, I guess we never do anything with the return value of `core/stop`, but I wanted to make sure we returned the same thing the original would.

PR #179, opened 2017-03-13
(maint) logging improvements

wkalt:
Add a bit more logging around controller disconnection/reconnection.


wkalt:
@MikaelSmith not pressing, but I found these additions helpful while working on ORCH-1583

PR #181, opened 2017-03-13
(ORCH-1601) add connection limit

wkalt:
Add a config setting to limit the number of client connections allowed
to the broker. Defaults to 0 (unlimited).

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -80,17 +80,19 @@ web-router-service: {
 }
 ```
 
-The broker can also be configured to make out-bound connections to other PCP servers. The
-connection is configured by providing a list of Websocket URIs, and optionally a whitelist
-of message types allowed to those connections. If the whitelist is not specified, only
-inventory requests are authorized. Note that all URIs must use a distinct hostname/port
-combination (the path is not taken into account when generating PCP identifiers).
+The broker exposes several configuration options around controller and client
+connections in the `pcp-broker` section. These options are:
 
-Users may also supply a controller disconnection graceperiod, which represents
-the amount of time to wait after all controllers become disconnected before
-evicting all client connections (to allow them to redistribute to other
-brokers). If unspecified, this defaults to 45 seconds. The parameter is
-supplied in milliseconds.
+* controller-uris: An array of Websocket URIs to which the broker will attempt
+  to establish outbound connections.
+* controller-whitelist: An array of message types the broker will accept from
+  connected controllers. Defaults to accepting only inventory requests.
+* controller-disconnection-graceperiod: The number of milliseconds after losing
+  connectivity to all configured controllers that the broker will wait before
+  dropping all connected clients (to allow them to redistribute to other
+  brokers).
+* max-connections: The maximum number of clients that can connect to the
+  broker. Defaults to 0, which is interpreted as unlimited.
 
 ```
 pcp-broker: {
@@ -98,5 +100,6 @@ pcp-broker: {
     controller-whitelist: ["http://puppetlabs.com/inventory_request",
                            "http://puppetlabs.com/rpc_blocking_request"],
     controller-disconnection-graceperiod: 45000
+    max-connections: 10000
 }
 ```
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -432,7 +432,7 @@
    Broker service is not running or if the client common name is not obtainable
    from its cert. Otherwise set the idle timeout of the WebSocket connection
    to 15 min."
-  [broker codec ws]
+  [{:keys [max-connections] :as broker} codec ws]
   (time!
    (:on-connect (:metrics broker))
    (cond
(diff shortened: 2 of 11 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -449,6 +451,9 @@
      (all-controllers-disconnected? broker)
      (websockets-client/close! ws 1011 (i18n/trs "All controllers disconnected"))
 
+     (and (pos? max-connections) (> @(:connection-count broker) max-connections))
+     (websockets-client/close!  ws 1011 (i18n/trs "Connection limit exceeded"))
```
wkalt: One thing we'll want to consider as the expectation shifts to load-balanced brokers is whether we should be adding the broker hostname to all of these close messages. Currently the broker-name isn't available here because this broker comes from init and broker-name isn't added until start, but I doubt it would be a hard change to make. I'm also not sure what our API commitments are to the message text.

doc/configuration.md
```diff
@@ -80,23 +80,26 @@ web-router-service: {
...
-
-Users may also supply a controller disconnection graceperiod, which represents
-the amount of time to wait after all controllers become disconnected before
-evicting all client connections (to allow them to redistribute to other
-brokers). If unspecified, this defaults to 45 seconds. The parameter is
-supplied in milliseconds.
+The broker exposes several configuration options around controller and client
+connections in the `pcp-broker` section. These options are:
+
+* controller-uris: An array of Websocket URIS to which the broker will attempt
```
MikaelSmith: `s` in URIS should be lower-case.

doc/configuration.md
```diff
@@ -80,23 +80,26 @@ web-router-service: {
...
+  to establish outbound connections.
+* controller-whitelist: An array of message types the broker will accept from
```
MikaelSmith: I'd add `defaults to accept only inventory requests`.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -432,9 +433,10 @@
    Broker service is not running or if the client common name is not obtainable
    from its cert. Otherwise set the idle timeout of the WebSocket connection
    to 15 min."
-  [broker codec ws]
+  [{:keys [max-connections] :as broker} codec ws]
   (time!
    (:on-connect (:metrics broker))
+   (swap! (:connection-count broker) inc)
```
MikaelSmith: Why use a separate counter, rather than size of inventory?

doc/configuration.md
MikaelSmith: Seems worth a comment at least. I can imagine it's so that we don't skip the check when many connections are being established, and we check before they're added to inventory.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: That seems useful.

src/puppetlabs/pcp/broker/service.clj
```diff
@@ -49,7 +52,9 @@
                                deref
                                :ssl-context-factory
                                .getSslContext)
-               broker (assoc broker :broker-name broker-name)
+               broker (assoc broker
+                             :broker-name broker-name
+                             :max-connections max-connections)
```
MikaelSmith: Why does this need to happen? Config shouldn't change between `init` and `start`.

wkalt: yeah, I switched when I realized the ws handlers came from init and didn't fix this. I'll remove it.

src/puppetlabs/pcp/broker/core.clj
wkalt: size of inventory seems better to me actually -- is there some other appeal to the connection count? shouldn't make a difference perf-wise I think.

MikaelSmith: I think size of inventory is fine, we should check for the limit close to where we add to inventory however.

doc/configuration.md
wkalt: @MichaelSmith this comment seems to relate to a different line?

MikaelSmith: The controller-whitelist, by default, will accept inventory requests.

src/puppetlabs/pcp/broker/core.clj
wkalt: I'm a little ambivalent about changing the call location -- by moving it further down, we end up doing extra work around authorization which we'll throw away. The only time we benefit from checking farther down afaict is when it's a superceded connection, and it's not clear to me what the right way to handle that is when the limit is exceeded either -- what if someone else connects after the superceded session was closed but before the new one is in inventory, etc.

It seems like the case where this limit is most likely to come into play is when one or more broker goes down and the others get flooded with connection requests, in excess of their limits. In that case I think we want to be closing the incoming connections as early as possible, even if it costs us being off by one sometimes when we're right at the limit.

doc/configuration.md
wkalt: I meant the one that says
```
Seems worth a comment at least. I can imagine it's so that we don't skip the check when many connections are being established, and we check before they're added to inventory.
```
I don't see how that relates to the defaults.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: Makes sense.

doc/configuration.md
MikaelSmith: Ah, I probably did put that in the wrong place. Seems related to the discussion around where we check the connection count.

wkalt:
@MikaelSmith fixed it up

PR #184, opened 2017-03-23
(PE-19981) Bump to clj-parent 0.6.1

rlinehan:
Bump to clj-parent 0.6.1 which includes trapperkeeper-metrics 1.0.0.

No code changes are required for this update.

The `:enabled` setting in the test metrics config was removed as this setting
had been deprecated and acted as a no-op for a long time; tk-metric 1.0.0
removed it entirely.

The metrics.conf included in packaging was updated to match the new config for
tk-metrics 1.0. As with the previous metrics.conf, it enables the JMX metrics
reporter. It also includes commented out example config for configuring the
Graphite reporter. The updates to this config file are not expected to cause
any problems because 1) it was never configurable using the PE module, and 2)
there is no documentation for how to configure any settings in it.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -13,7 +13,7 @@
   ;; requires lein 2.2.0+.
   :pedantic? :abort
 
-  :parent-project {:coords [puppetlabs/clj-parent "0.6.0"]
+  :parent-project {:coords [puppetlabs/clj-parent "0.6.1"]
                    :inherit [:managed-dependencies]}
 
   :dependencies [[org.clojure/clojure]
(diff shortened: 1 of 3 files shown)


rlinehan:
ping @nicklewis here's the PR that updates the config file in packaging (and mentions the PE module and that changes are not needed).

MikaelSmith:
Tests pass but running `lein tk` fails. This is used in pxp-agent tests (I thought we'd stopped using it on pxp-agent#stable, but I appear to be mistaken) so reverting for now.

PR #190, opened 2017-05-05
(PCP-741) Upgrade jetty to 9.4

adreyer:
blocked on https://github.com/puppetlabs/clj-pcp-client/pull/77

This changed the method of instrospecting brokername from the ssl
context to use only public methods.

### DIFF
--- a/test/utils/puppetlabs/pcp/testutils/server.clj
+++ b/test/utils/puppetlabs/pcp/testutils/server.clj
@@ -52,5 +52,10 @@
 
 (defn wait-for-inbound-connection
   [svc-context]
-  (while (empty? @(:inventory svc-context))
-    (Thread/sleep 100)))
+  (loop [i 0]
+    (when (empty? @(:inventory svc-context))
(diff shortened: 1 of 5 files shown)


test/utils/puppetlabs/pcp/testutils/server.clj
```diff
@@ -52,5 +52,10 @@
 
 (defn wait-for-inbound-connection
   [svc-context]
-  (while (empty? @(:inventory svc-context))
-    (Thread/sleep 100)))
+  (loop [i 0]
```
MikaelSmith: What led to this change? You didn't want an infinite loop?

MikaelSmith:
We should bump to 1.4.0-SNAPSHOT for this.

test/utils/puppetlabs/pcp/testutils/server.clj
adreyer: Yeah, the bug in error handling was causing an infinite loop in some tests.

PR #191, opened 2017-05-08
(SEC-36) Enabling static Clojure code scans using a shell script

jsane:
We (security team) is planning to enable/start static code scans on clojure. This is a first small step and currently it only looks for [clojure.core] read-string instances. The grep in the script is not context aware and so can generate false positives if read-eval has been disabled either in local scope or globally which the script cannot detect. 
That apart it should not yield any false positives and neither should there be any false negatives. Although no read-strings are being used currently this is to guard against any inadvertent use in future.

If you suspect or notice any issue with it interfering with the build, feel free to disable it by commenting the script invocation from .travis.yml. But do drop a note to me and/or Bill about the failure so we can rectify.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,7 @@ jdk:
 script:
   - lein test
   - lein with-profile test-schema-validation test
+  - chmod +x $TRAVIS_BUILD_DIR/ext/travisci/secscan.sh
+  - $TRAVIS_BUILD_DIR/ext/travisci/secscan.sh "$TRAVIS_BUILD_DIR/src" "clj" "clojure.core/read-string"
 notifications:
   email: false
--- a/ext/travisci/secscan.sh
+++ b/ext/travisci/secscan.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+#set -e
+
+
+# TODO::
+# . It would also be nice if it could be passed the pattern as an argument but that may not be feasible 
+#    when one considers substring matches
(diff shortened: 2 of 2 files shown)


jsane:
Retracting the PR as it did not follow the convention of creating a branch for the issue in the fork. Also the build seems to fail though it is not due to the changes introduced so need to discuss how to enable the scans while overcoming the (existing) build failures in travis.

jsane:
@MikaelSmith I am thinking of re-submitting by taking a proper route (creating an issue branch  on my fork etc.) like I indicated above. I also have a few minor nits/fixes in the shell script that I would like to bring in. 
Regarding Travis CI builds: Not sure if this the right place (https://travis-ci.com/puppetlabs/) to look for all travis ci builds for puppetlabs repo's but don't see pcp-broker on it.

MikaelSmith:
It's a public repo, so it'll be under https://travis-ci.org/puppetlabs.

Also, the branch name is fine as-is unless you really want to change it. I'll just have to massage the merge commit text.

jsane:
Ahh.. right. About the build freezing up on my fork (yes I did check this on travis-ci.org) it seems to be a temporary thing as I checked today earlier it had passed fine. Maybe Travis gives that as an interim warning about it taking long etc.  
In any case am re-submitting the PR from a different branch (it also has those minor nits fixed in the shell script). Will keep the PR merge comment brief so you can adjust appropriately. Hope that is ok. thanks.

MikaelSmith:
Sounds good.

PR #196, opened 2017-06-06
(PCP-732) Refine logging in broker

mruzicka:
This PR updates the log levels of the logged messages according to the following definition:
ERROR: abnormal failure to startup, establish a connection, or deliver a message
WARNING: authorization or authentication failure, or failure to process a message
INFO: change in service or connections (connect/disconnect)
DEBUG: sending or responding to messages with identifying information (but not content), completing state transitions (namely those logged at INFO)
TRACE: receiving messages with full contents, full contents of responses

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -106,16 +106,16 @@
     (when (not= type (ws->client-type (:websocket connection)))
       (let [{:keys [uri]} connection]
         (sl/maplog
-         :debug (assoc (connection/summarize connection)
-                       :uri as
-                       :existinguri uri
-                       :type :connection-already-associated)
+         :info (assoc (connection/summarize connection)
+                      :uri as
+                      :existinguri uri
+                      :type :connection-already-associated)
          ;; Logs association status with the following arguments
          ;; 0 : uri associated with incoming connection
          ;; 1 : certificate common name of connection
          ;; 2 : remote address of connection
          ;; 3 : uri previously associated with the connection
-         #(i18n/trs "Received session association for {0} from {1} {2}. Session was already associated as {3}"
+         #(i18n/trs "Rejecting session association for {0} from {1} {2}. Session was already associated as {3}"
            (:uri %) (:commonname %) (:remoteaddress %) (:existinguri %)))
         (i18n/trs "Session already associated")))))
 
@@ -165,15 +165,16 @@
        (locking (:websocket connection)
          (send-message connection message))
        (catch Exception e
-         (sl/maplog :debug e
-                    {:type :message-delivery-error}
-                    (fn [_] (i18n/trs "Error in process-associate-request!")))))
+         (sl/maplog :warn e
+                    {:target requester-uri
+                     :type :message-delivery-error}
+                    #(i18n/trs "Attempted associate response delivery to {0} failed." (:target %)))))
      (if reason-to-deny
        (do
          (sl/maplog
-          :debug {:type   :connection-association-failed
-                  :uri    requester-uri
-                  :reason reason-to-deny}
+          :warn {:type   :connection-association-failed
+                 :uri    requester-uri
+                 :reason reason-to-deny}
           ;; 0 : reason to deny request
           ;; 1 : uri of connection
           #(i18n/trs "Invalid associate_request ({0}); closing {1} WebSocket" (:reason %) (:uri %)))
@@ -218,9 +219,9 @@
       "http://puppetlabs.com/associate_request" (process-associate-request! broker message connection)
       "http://puppetlabs.com/inventory_request" (process-inventory-request broker message connection)
       (sl/maplog
-       :debug (assoc (connection/summarize connection)
-                     :messagetype message-type
-                     :type :broker-unhandled-message)
+       :warn (assoc (connection/summarize connection)
+                    :messagetype message-type
+                    :type :broker-unhandled-message)
        ;; 0 : message type
        ;; 1 : common name of connection
        ;; 2 : remote address of connection
@@ -396,9 +397,9 @@
           (assert false (i18n/trs "unexpected message validation outcome")))
         (catch map? m
           (sl/maplog
-           :debug (merge (connection/summarize connection)
-                         (summarize message)
-                         {:type :processing-error :errortype (:type m)})
+           :warn (merge (connection/summarize connection)
+                        (summarize message)
+                        {:type :processing-error :errortype (:type m)})
            ;; Failure while reading the message.
            ;; 0 : message type
            ;; 1 : message id (uuid)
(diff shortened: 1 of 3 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -165,15 +165,15 @@
        (locking (:websocket connection)
          (send-message connection message))
        (catch Exception e
-         (sl/maplog :debug e
+         (sl/maplog :error e
```
MikaelSmith: This message happens because the connection is no longer open. Maybe we should make this info, and change the message to something like "Unable to deliver association response because connection is no longer available.".

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -106,10 +106,10 @@
...
-                       :uri as
-                       :existinguri uri
-                       :type :connection-already-associated)
+         :info (assoc (connection/summarize connection)
+                      :uri as
+                      :existinguri uri
+                      :type :connection-already-associated)
          ;; Logs association status with the following arguments
          ;; 0 : uri associated with incoming connection
          ;; 1 : certificate common name of connection
```
MikaelSmith: I think we should modify this message a bit to be clear what was done in response: the connection was rejected. So maybe "Rejecting session association for {0} from {1} {2}. Session was already associated as {3}."

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -644,11 +644,11 @@
    client :- Client]
   (let [timestamp (now)]
     (sl/maplog
-      :debug {:uri uri}
+      :info {:uri uri}
       ;; 0 : connection uri
       #(i18n/trs "Lost connection to controller: {0}" (:uri %)))
     (sl/maplog
-      :debug {:uri uri}
+      :info {:uri uri}
```
MikaelSmith: Having these two messages one after the other seems odd. I'm not sure the 2nd (removing inventory update subscription) is necessary. We don't reference inventory update subscription anywhere else.

src/puppetlabs/pcp/broker/shared.clj
```diff
@@ -116,16 +116,16 @@
       (locking (:websocket connection)
         (send-message connection error-msg))
       (catch Exception e
-        (sl/maplog :debug e
+        (sl/maplog :error e
```
MikaelSmith: This is another case like the error in `process-associate-request!`. I'd be tempted to call this `info`, and change the message to be more clear, something like "Attempted message delivery failed as connection is no longer available."

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: It'd also be nice to incorporate the `requester-uri` in the log message.

src/puppetlabs/pcp/broker/shared.clj
MikaelSmith: It'd also be nice to incorporate the `requester-uri` in the log message.

mruzicka:
@MikaelSmith I've addressed your comments, the travis build failed on the first run, on a re-run it succeeded, the error was not related to the changes in this PR.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -165,15 +165,16 @@
        (locking (:websocket connection)
          (send-message connection message))
        (catch Exception e
-         (sl/maplog :debug e
-                    {:type :message-delivery-error}
-                    (fn [_] (i18n/trs "Error in process-associate-request!")))))
+         (sl/maplog :info e
+                    {:type :message-delivery-error
+                     :target requester-uri}
+                    #(i18n/trs "Error sending associate response to {0}" (:target %)))))
```
MikaelSmith: It feels incongruous calling this `info` and still saying "Error". Was there something you didn't like about the phrasing I suggested?

src/puppetlabs/pcp/broker/shared.clj
```diff
@@ -116,16 +116,17 @@
       (locking (:websocket connection)
         (send-message connection error-msg))
       (catch Exception e
-        (sl/maplog :debug e
-                   {:type :message-delivery-error}
-                   (fn [_] (i18n/trs "Error in send-error-message")))))))
+        (sl/maplog :info e
+                   {:type :message-delivery-error
+                    :target (:uri connection)}
+                   #(i18n/trs "Error sending error message to {0}" (:target %)))))))
```
MikaelSmith: As above, I suggested some phrasing that you appear to have ignored. `info` and "Error" are incongruous. I'd like to make that clearer somehow.

MikaelSmith:
Similar to the two outstanding suggestions, there are two errors in shared.clj that say "Error in deliver-message" that I think we should clarify (and maybe downgrade to `info`).

mruzicka:
@MikaelSmith I've updated the messages according to your suggestion but left out the "connection is no longer available" part as IMO that need not be always the case.
I've also changed the level of the log entries of the four message delivery problems to `:warn`.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -633,7 +634,7 @@
   (future (do (Thread/sleep controller-disconnection-ms)
               (maybe-purge-clients! broker timestamp)))
   (sl/maplog
-    :debug {:timeout controller-disconnection-ms}
+    :info {:timeout controller-disconnection-ms}
```
adreyer: Is this going to make people think things are purged even though they may not be?

mruzicka:
I'd go as far as logging the no client certificate case at the warning level as it is an "authorization or authentication failure".

src/puppetlabs/pcp/broker/core.clj
mruzicka: I've changed the message to say "potential eviction" and added another log to the function where the eviction actually happens.

mruzicka:
@ipeldridge I've updated the messages concerning the client eviction.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -508,9 +509,9 @@
              (websockets-client/close! (:websocket old-conn) 4000 (i18n/trs "superseded")))
```
mruzicka: This is not even logged. It's sent to the connection peer as the reason for closing the connection.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -635,7 +639,7 @@
   (sl/maplog
     :debug {:timeout controller-disconnection-ms}
     ;; 0 : number of milliseconds
-    #(i18n/trs "Scheduled full client eviction in {0,number,integer} ms" (:timeout %))))
+    #(i18n/trs "Scheduled potential full client eviction in {0,number,integer} ms" (:timeout %))))
```
mruzicka: I'd go with the abbreviation as we don't need to care about pluralizing it.

PR #197, opened 2017-06-16
(PCP-737) Handle events on stale connection gracefully

mruzicka:
This PR makes sure errors or messages received on stale websocket connections are handled gracefully. A stale connection in this context means a connection which is not present in the broker's inventory any longer.
Previously it was not the case which could lead to validation errors when there was an attempt to log the connection details for connection which was not found in the broker's inventory.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -7,7 +7,7 @@
             [puppetlabs.pcp.broker.metrics :as metrics]
             [puppetlabs.pcp.broker.message :as message :refer [Message multicast-message?]]
             [puppetlabs.pcp.broker.inventory :as inventory]
-            [puppetlabs.pcp.broker.util :refer [assoc-when]]
+            [puppetlabs.pcp.broker.util :refer [assoc-when] :as util]
             [puppetlabs.pcp.client :as pcp-client]
             [puppetlabs.pcp.protocol :as p]
             [puppetlabs.metrics :refer [time!]]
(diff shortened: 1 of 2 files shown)


mruzicka:
A question to reviewers: It seems possible we could see similar problem at other places in the broker code e.g. [here](https://github.com/mruzicka/pcp-broker/blob/376ff12e3e4f4c1efee1a40c8b810b4b8261130f/src/puppetlabs/pcp/broker/core.clj#L357) - should we handle the eventuality there too (although I can imagine it should be much less likely that we actually see the problem there in comparison to the on-error handler) or is throwing an exception there appropriate?

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -518,9 +518,9 @@
                         (:uri %) (:remoteaddress %)))))))))
 
 (defn- on-error
-  "OnError WebSocket event handler. Just log the event."
+  "OnError WebSocket event handler. Just log the event if the connection is active that is."
```
MikaelSmith: Sentence reads a bit funny. Maybe rephrase to "Log the event if the connection is active."

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -518,9 +518,9 @@
...
   [broker ws e]
-  (let [connection (get-connection broker (ws->uri ws))]
+  (if-let [connection (get-connection broker (ws->uri ws))]
```
MikaelSmith: I think I'd like to also log if the connection is inactive, we just can't include as much info in the message.

MikaelSmith:
Yeah, we probably should handle that. Maybe just make the summarize argument optionally nil and return empty data?

mruzicka:
I've changed the PR to log not ignore errors on stale connections but rather log it at debug level, I've also added logging of messaged received on stale connections.
The PR is probably easier to review using [?w=1](197/files?w=1).

PR #199, opened 2017-06-22
(PCP-737) Handle events on stale controller connections too

mruzicka:
A recent PR made sure events on stale connection are handled gracefully. (A stale connection meaning a connection which is not present in the broker's inventory any longer.) But it failed to address similar problem on controller connections which are kept in a separate inventory / registry.
So this PR makes sure even events on stale controller connections are handled gracefully.

Not that it is easier to review using [?w=1](199/files?w=1)

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -590,39 +590,48 @@
    whitelist :- #{s/Str}
    client :- Client
    message :- Message]
-  (let [uri (ws->uri client)
-        connection (get-controller broker uri)
-        message (assoc-when message :sender uri :target "pcp:///server")
-        message-data (merge (connection/summarize connection) (summarize message))]
-    (time!
-     (:on-message (:metrics broker))
+  (let [uri (ws->uri client)]
+    (if-let [connection (get-controller broker uri)]
+      (let [message (assoc-when message :sender uri :target "pcp:///server")
+            message-data (merge (connection/summarize connection) (summarize message))]
+        (time!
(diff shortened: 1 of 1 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -590,39 +590,48 @@
...
-        :else
-        (do
-          (log-access :info (assoc message-data :accessoutcome "AUTHORIZATION_SUCCESS"))
-          (if (= (:target message) "pcp:///server")
-            (process-server-message! broker message connection)
-            (deliver-message broker message connection)))))))
+                 ;; 0 : uri of connection
+                 ;; 1 : remote address of connection
+                 ;; 2 : raw message
+                 #(i18n/trs "Ignoring message received on stale controller connection {0} {1}: {2}"
```
MikaelSmith: Formatting's slightly inconsistent with the other location. Should we also say "Message dump: " here?

MikaelSmith: I guess the result is slightly different, hex vs message.

mruzicka: We can't decode a message received on a stale agent connection since we don't know the codec to use to decode the messages as it is an attribute of the connection record which is obviously not available, that's why I opted for the hxdump in that case.
On the other hand the controller connections don't use the concept of codecs and therefore the messages received on the controller connections can be decoded even when the connections are stale i.e. it is always possible to log the messages from controllers in their decoded forms which I took advantage of in this case.

PR #201, opened 2017-06-28
(PCP-735) Allow disconnect to resolve

MikaelSmith:
Introduce a small wait after receiving the on-close message to ensure
the websocket object registers that the connection is closed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,10 @@ lein: 2.7.1
 jdk:
   - oraclejdk8
 script:
-  - lein test
-  - lein with-profile test-schema-validation test
-  - chmod +x $TRAVIS_BUILD_DIR/ext/travisci/secscan.sh
-  - $TRAVIS_BUILD_DIR/ext/travisci/secscan.sh "$TRAVIS_BUILD_DIR/src" "clj" "clojure.core/read-string"
+  - lein $TEST_OPTIONS test
+  - if [ "$TEST_OPTIONS" = "" ]; then ./ext/travisci/secscan.sh src clj "clojure.core/read-string"; fi
+env:
+  - TEST_OPTIONS=""
+  - TEST_OPTIONS="with-profile test-schema-validation"
 notifications:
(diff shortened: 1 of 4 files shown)


test/integration/puppetlabs/pcp/broker/controllers_test.clj
```diff
@@ -57,31 +57,47 @@
                       :target agent-uri
                       :data "Hello"}))
 
+(defn verify-response
+  [response message_type message expected]
```
mruzicka: I'm not a huge fan of factoring out of assertions into functions called from different places as it makes running/debugging of the tests in IntelliJ IDE less convenient / transparent.
But that's just my preference ...

test/integration/puppetlabs/pcp/broker/controllers_test.clj
```diff
@@ -57,31 +57,47 @@
                       :target agent-uri
                       :data "Hello"}))
 
+(defn verify-response
```
mruzicka: I'm not a huge fan of factoring out of assertions into functions called from different places as it makes debugging of the tests in IntelliJ IDE less convenient / transparent.
But that's just my preference ...

MikaelSmith: Do you know any way to reuse code? I have the same dislike as well, but couldn't come up with something better. I guess I could remove that refactor.

MikaelSmith: Removed them.

.travis.yml
```diff
@@ -4,9 +4,10 @@ lein: 2.7.1
 jdk:
   - oraclejdk8
 script:
-  - lein test
-  - lein with-profile test-schema-validation test
-  - chmod +x $TRAVIS_BUILD_DIR/ext/travisci/secscan.sh
```
mruzicka: This was discarded, is it not necessary?

PR #204, opened 2017-08-22
(PCP-777) Reject connections before websocket upgrade

MikaelSmith:
Accepting websocket upgrades before rejecting the connection causes
clients to get a successful connection before it immediately disappears.
It's difficult to distinguish that this is a configuration issue - vs a
broker going away - to determine whether the client should immediately
attempt to reconnect and get a new broker from the load balancer, or
delay reconnects.

Update the broker to reject connections immediately - by disabling the
routing handler - rather than accepting a websocket upgrade before
disconnecting.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -459,6 +459,28 @@
        (= (set (keys @(:controllers broker)))
           (set (keys (:warning-bin @(:database broker)))))))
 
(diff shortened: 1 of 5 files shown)


nicklewis:
This change looks good to me, but I want to do some thorough testing to make sure it's free of race conditions.

nicklewis:
I tried this out a bit and it seems fine. I'm less concerned about it than I was earlier, after remembering there's a 90 second grace period before agents get dropped / handlers get closed, which means the possibility of this flapping wildly isn't realistic.

PR #205, opened 2017-08-22
(PCP-491) Make max message size configurable

MikaelSmith:
Adds a `pcp-broker.max-message-size` configuration option that
configures the max binary message size for WebSocket sessions. This
would allow large messages to be sent by clients.


nicklewis:
This doesn't seem to solve the problem I was seeing...

MikaelSmith:
We should figure that out.

nicklewis:
The commit message is wrong now (mentions binary).

MikaelSmith:
Fixed.

PR #209, opened 2018-01-03
(maint) Use artifactory instead of nexus

sarameisburger:
We are making the change from nexus to artifactory


sarameisburger:
Will be merged up. Closing.

PR #217, opened 2019-02-21
(PCP-862) Use disconnect rather than close

lucywyman:
This changes closing the pcp connection behavior to use disconnect in place of close, in hopes that disconnect will stop the connection more quickly and reduce the number of file descriptors opened by pcp at any given time.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -13,7 +13,7 @@
   ;; requires lein 2.2.0+.
   :pedantic? :abort
 
-  :parent-project {:coords [puppetlabs/clj-parent "2.0.0"]
+  :parent-project {:coords [puppetlabs/clj-parent "2.6.4"]
                    :inherit [:managed-dependencies]}
 
   :dependencies [[org.clojure/clojure]
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -537,7 +537,8 @@
                         ;; 2 : remote address of connection
                         #(i18n/trs "Node with URI {0} already associated with connection {1} {2}."
                           (:uri %) (:commonname %) (:remoteaddress %)))
-             (websockets-client/close! (:websocket old-conn) 4000 (i18n/trs "Superseded.")))
+             (websockets-client/close! (:websocket old-conn) 4000 (i18n/trs "Superseded."))
+             (websockets-client/disconnect (:websocket old-conn)))
            (websockets-client/idle-timeout! ws (* 1000 60 15))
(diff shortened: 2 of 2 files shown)


src/puppetlabs/pcp/broker/core.clj
```diff
@@ -537,7 +537,7 @@
                         ;; 2 : remote address of connection
                         #(i18n/trs "Node with URI {0} already associated with connection {1} {2}."
                           (:uri %) (:commonname %) (:remoteaddress %)))
-             (websockets-client/close! (:websocket old-conn) 4000 (i18n/trs "Superseded.")))
```
MikaelSmith: Part of how we identified an issue was the `Superseded` messages. I'd like to preserve that somehow. Did `close!` then `disconnect` not work?

Maybe the log message is enough.

project.clj
```diff
@@ -22,7 +22,7 @@
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/trapperkeeper-authorization]
                  [puppetlabs/trapperkeeper-metrics]
-                 [puppetlabs/trapperkeeper-webserver-jetty9]
+                 [puppetlabs/trapperkeeper-webserver-jetty9 "2.4.0"]
```
MikaelSmith: We'll want to update this in clj-parent sooner rather than later.

MikaelSmith: And we'll have to get the right version in pe-orchestration-services and pe-puppetserver.

src/puppetlabs/pcp/broker/core.clj
MikaelSmith: If this works as effectively as our testing with just disconnect, 👍

project.clj
lucywyman: Ahh, I didn't realize that's where it updated! Thanks!

lucywyman:
https://github.com/puppetlabs/pcp-broker/blob/master/test/integration/puppetlabs/pcp/broker/subscription_test.clj#L157 This is the line that's hanging. Still trying to figure out why....

MikaelSmith:
Could that test somehow have been superseded by another connection?

lucywyman:
https://github.com/eclipse/paho.mqtt.java/issues/330#issuecomment-448568449 I think this is what's happening? The connection hasn't been fully established, so when it tries to disconnect it hangs

MikaelSmith:
Why would hang during disconnect cause it to hang at `deref @subscription-finished`?

lucywyman:
:tada:

PR #218, opened 2019-02-28
(PCP-862) Remove `close!` call before disconnecting

lucywyman:
This removes the `close!` call before disconnecting a superseded connection. It was originally left to preserve the code and reason in logging, but appeared to leave unused sockets open, consuming resources even when disconnect was called. Using just `disconnect` frees the resources, and we determined there was already sufficient logging of pxp-agents being superseded to lose the 'Superseded' message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.5.3
+
+This is a maintenance release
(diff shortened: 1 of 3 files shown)


MikaelSmith:
I think we should include something about Close leaving unused sockets open as a rationale for preferring disconnect over close. I wish I had a clearer explanation though.

PR #222, opened 2019-08-28
(maint) Add bouncycastle to integration profile

branan:
This adds the bouncycastle dependency to the `:integration` profile just as it is part of the `:dev` profile. It also creates a new `:internal-integration` which allows the `:integration` profile to be used alongisde the mirrors specified in `:internal-mirrors`.


jonathannewman:
LGTM! Yes

PR #227, opened 2020-07-13
(maint) Allow adjusting the idle timeout and lower its default

justinstoller:
pcp-broker appears to have hang on to SSLContexts and connections longer than necessary in dynamic environments where agents frequently come and go. While idle-timeout is most likely not contributing significantly to that, it seems like a simple thing to lower and help in a small way.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -490,7 +490,7 @@
    Broker service is not running or if the client common name is not obtainable
    from its cert. Otherwise set the idle timeout of the WebSocket connection
    to 15 min."
-  [{:keys [max-connections] :as broker} codec ws]
+  [{:keys [max-connections idle-timeout] :as broker} codec ws]
   (time!
    (:on-connect (:metrics broker))
    (cond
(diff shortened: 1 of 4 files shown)


MikaelSmith:
I assume the agents are being shutdown in a way where they don't close their TCP connections?

justinstoller:
I'm unclear on how exactly their connections could be orphaned but the logs I looked at were seeing 5-20 ish timeouts a day.

MikaelSmith:
Ok. The only thing this might interact strangely with is pxp-agent's timeout period (which is 2-3x the ping interval that defaults to 2 minutes). I can't think of a reason that would specifically cause issues. @nicklewis do you recall anything?

nicklewis:
This seems fine. I think the stipulations are that the pcp-broker's idle timeout **must** be longer than the pxp-agent ping interval (or agents will continually timeout) and **should** be longer than the total timeout period (or the broker will tolerate fewer missed pings than the agents is configured to allow). Otherwise it should be okay.

justinstoller:
Okay, it sounds like I should configure this default to be 6 minutes so the agent and broker will both give up on a connection in similar timeframes. I'll also mention PCP-729 a commit message since it looks like that's the ticket that led us to the current agent timeouts.

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -779,10 +779,12 @@
                 get-route
                 get-metrics-registry
                 max-connections
-                max-message-size]} options
+                max-message-size
+                idle-timeout]} options
         broker  {:broker-name         broker-name
                  :max-connections     max-connections
                  :max-message-size    max-message-size
+                 :idle-timeout        idle-timeout
```
MikaelSmith, line 787: It looks like this also requires a change to the schema in shared.clj.

PR #228, opened 2020-07-13
(PCP-900) Close connections after CRL update

justinstoller:
tk-jetty9 will reload the SSLContextFactory for pcp-broker when changes
to the CRL occur. In doing so new connections will use the CRL updates.
However pcp-broker, by design, keeps connections open for long durations.
This can result in many different SSLContexts in the heap with different
versions of the CRL. Coupled with large CRLs this can take up a
significant portion of the heap (10-20%).

This brings in tk-filesystem-watcher as a dependency and, like
tk-jetty9, watches the CRL for updates. When it has changed the client
are disconnected fromt the server. This is under the assumption that the
clients will reconnect and the server will be able to use the new CRL.

Some caveats:

 - The SSLContextFactory is not guaranteed to have reloaded by the time
   the first agent reconnects. This might cause 2 CRL versions in memory,
   but should limit the number to 2.  Unfortunately, even though
   SSLContextFactory implements a Lifecycle listener pattern, reloads do
   not seem to trigger them (the code looks like only start and stop are
   implemented).

 - This could cause significant churn in large deployments.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -17,13 +17,16 @@
             [puppetlabs.trapperkeeper.authorization.ring :as ring]
             [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-core :as jetty9-core]
+            [puppetlabs.trapperkeeper.services.protocols.filesystem-watch-service :as watch]
             [schema.core :as s]
             [slingshot.slingshot :refer [throw+ try+]]
+            [me.raynes.fs :as fs]
             [puppetlabs.i18n.core :as i18n])
   (:import [puppetlabs.pcp.broker.connection Connection]
            [puppetlabs.pcp.client Client]
            [clojure.lang IFn]
            [java.net InetAddress UnknownHostException URI]
+           [org.eclipse.jetty.util.ssl SslContextFactory]
            [javax.net.ssl SSLContext]
            [java.security KeyStore]))
 
@@ -689,6 +692,18 @@
       :info {}
       (fn [_] (i18n/trs "Evicted all clients as there is no controller connection.")))))
 
+(defn restart-client-connections!
+  "Forces clients to reconnect because the broker's CRL is out of date"
+  [broker]
+  (doseq [[_ client-connection] (:inventory @(:database broker))]
+    (websockets-client/close!
+      (:websocket client-connection)
+      1012 ;; code SERVER_RESTART, client *should* reconnect soon
+      (i18n/trs "CRL reloaded"))
+    (sl/maplog
+      :info {}
+      (fn [_] (i18n/trs "Evicted all clients because of CRL reload")))))
+
 (defn on-controller-connect!
   [broker controller-uri ws]
   ;; lock to ensure that updating connected controllers and restarting handlers happen together
@@ -757,6 +772,23 @@
   (into {} (pmap (partial start-client broker ssl-context
                           controller-whitelist controller-disconnection-ms)
                  controller-uris)))
+
+(s/defn watch-crl
+  [watcher :- (s/protocol watch/Watcher)
+   broker :- Broker
+   ssl-context-factory :- SslContextFactory]
+  (when-let [crl-config-path (.getCrlPath ssl-context-factory)]
+    (let [crl-path (.getCanonicalPath (fs/file crl-config-path))]
+      (watch/add-watch-dir! watcher (fs/parent crl-path))
+      (watch/add-callback!
+       watcher
+       (fn [events]
+         (when (some #(and
+                       (:changed-path %)
+                       (= (.getCanonicalPath (:changed-path %))
+                          crl-path))
+                     events)
+           (restart-client-connections! broker)))))))
 
 ;;
 ;; Broker service lifecycle, status service
--- a/project.clj
+++ b/project.clj
@@ -24,6 +24,7 @@
                  [puppetlabs/trapperkeeper-metrics]
                  [puppetlabs/trapperkeeper-webserver-jetty9]
                  [puppetlabs/trapperkeeper-status]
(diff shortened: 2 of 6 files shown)


justinstoller:
Needs tests, but I want to get feedback on the approach and conclusions I've made.

MikaelSmith:
There's no way to identify specific clients that were added to the CRL and only disconnect them?

I'd also encourage spacing out the disconnects over several minutes so that reconnects go more smoothly.

justinstoller:
> There's no way to identify specific clients that were added to the CRL and only disconnect them?

The main issue I'm concerned about is the amount of heap dedicated to old CRLs (vs disconnection of potentially revoked clients). Jetty doesn't seem to do any incremental updates, when it "reloads" an SSLContextFactory it creates a new CRLImpl with new CRLEntries, the vast majority are duplicates of each other. The SSLContextFactory no longer holds a reference to the old CRLImpl and once the SSLContexts created from the factory are done it should be garbage collected. But afaict, the open Websocket connections are holding a reference to the SSLContext they were created with....

I'm not sure if the individual connections can renegotiate with a new SSLContext? I can look into that.

MikaelSmith:
Ah. So if we wont'd to do a rolling restart we'd have to switch all new connections over to a new SSLContext (if that's possible) and then disconnect them from the old one before shutting down. Yeah, not sure off the top of my head how feasible that is.

justinstoller:
I didn't see anything that would work here or its implementing classes: [AbstractWebSocketConnection](https://www.eclipse.org/jetty/javadoc/9.4.30.v20200611/org/eclipse/jetty/websocket/common/io/AbstractWebSocketConnection.html)

justinstoller:
Give that, does the general approach look sane?

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -689,6 +692,18 @@
       :info {}
       (fn [_] (i18n/trs "Evicted all clients as there is no controller connection.")))))
 
+(defn restart-client-connections!
+  "Forces clients to reconnect because the broker's CRL is out of date"
+  [broker]
+  (doseq [[_ client-connection] (:inventory @(:database broker))]
+    (websockets-client/close!
+      (:websocket client-connection)
+      1012 ;; server restart, client should reconnect in 5-30s
```
MikaelSmith, line 701: Where does that 5-30s come from?

justinstoller: I saw that mentioned in Jetty's docs here [SERVICE_RESTART](https://www.eclipse.org/jetty/javadoc/9.4.6.v20170531/org/eclipse/jetty/websocket/api/StatusCode.html#SERVICE_RESTART) Of those listed on that page it seemed like the best to send, I wasn't sure if there was any value in sending a custom code.

MikaelSmith: That makes sense, but I think the retry period is client-specific.

justinstoller: I'll remove that portion of the comment

MikaelSmith:
Does this relate to a ticket?

justinstoller:
I'll create a PCP ticket. This was part of an investigation for a customer in PE-27722, where we thought part of the issue was PCP-862 and after upgrade the situation (wrt many old CRLs in the heap) was improved but not fully resolved.

justinstoller:
When writing a test for this my test is always returning a 1006 close code. I don't see how what I'm calling is different than anything else at this point though. The test also seems relatively straightforward. I see in another test that says 1006 is racy and filters for them, however I've ran the assertion in a loop and seen it _only_ ever get 1006s. However in my manual testing I did see:
```
2020-07-13 05:33:05.423719 DEBUG puppetlabs.cpp_pcp_client.connection:545 - WebSocket on close event: Success (code: 1012) - connection timings: TCP 1412 us, WS handshake 72992 us, overall 79240 us
2020-07-13 05:33:05.423924 WARN  puppetlabs.cpp_pcp_client.connector:292 - WebSocket connection to PCP broker lost; retrying
2020-07-13 05:33:05.624104 DEBUG puppetlabs.cpp_pcp_client.connection:487 - WebSocket TLS initialization event; about to validate the certificate
2020-07-13 05:33:05.624616 DEBUG puppetlabs.cpp_pcp_client.connection:523 - Initialized SSL context to verify broker grand-checker.delivery.puppetlabs.net
....
```
I'm wondering if those who are more familiar with the project and its testing can let me know if there's something I should be doing differently for local testing and/or if the 1006 return code is something I should care about?

MikaelSmith:
I don't have any special context, https://github.com/puppetlabs/pcp-broker/commit/0b707aeb3ac86dfb74edea12d51f533624749f14 introduced the workaround.

justinstoller:
I opened this https://github.com/puppetlabs/pcp-broker/pull/230 which shows master failing on java 8 in a similar way. It looks like there may be some recent change in Java8 that no longer allows CA certs with incorrect extensions.

justinstoller:
Java 11 is failing with my recent test for this change. I haven't been able to replicate the issue locally, but still working on it.

justinstoller:
Per https://github.com/puppetlabs/pcp-broker/pull/228#issuecomment-663777969 and #230 do you want me to try and rebuild the certs in this repo?

MikaelSmith:
Seems like that's necessary to get CI passing again, I'm in favor.

justinstoller:
Okay, rebased and that fixed the Java 8 issues. However, Java 11 was still failing (but I couldn't reproduce locally). Did some experiments and it seems like on Java 11 in travis the CRL was being reloaded on the server before the client finished establishing its connection. I've put a simple sleep in there because I didn't want to change the arity of `connect-and-close`. I went with not updating `connect-and-close` because in my experiments I made that helper slower by printing out everything it received and that caused other tests to start failing sporadically.

I could make a separate helper that is very similar to connect and close if folks want me to remove the sleep statement. Otherwise I think this is finally ready to merge.

PR #233, opened 2020-10-13
(PE-30041) Remove harmful term from pcp-broker config

donoghuc:
This commit changes instances of `whitelist` to `allowlist`.

### DIFF
--- a/doc/authorization.md
+++ b/doc/authorization.md
@@ -4,7 +4,7 @@ Messages from clients can be allowed or denied using the trapperkeeper-authoriza
 
 https://github.com/puppetlabs/trapperkeeper-authorization/blob/master/doc/authorization-config.md
 
-Messages from controller out-bound connections are authorized via a `controller-whitelist`
+Messages from controller out-bound connections are authorized via a `controller-allowlist`
 configuration option. See [configuration](configuration.md) for details.
 
 ## Message mapping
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -131,7 +131,7 @@ The major updates are:
   inventory updates, which were introduced in the [PCP v2
   specifications](https://github.com/puppetlabs/pcp-specifications/blob/master/pcp/versions/2.0/inventory.md#inventory-messages)
 * [PCP-681](https://tickets.puppetlabs.com/browse/PCP-681) Add the
-  `controller-uris`  and `controller-whitelist` configuration options. These
+  `controller-uris`  and `controller-allowlist` configuration options. These
```
steveax, line 134: Seems like we should not retroactively change that note, but rather add a new change to the Changelog detailing the name change?

steveax (commented):
Just one nit on the changelog. It's not 100% clear to me if we need to allow for the legacy setting name though. Will an upgrade take care of remapping the configuration keys?

barriserloth (approved):
I'm +1 on this, but should we add a note about the change to the changelog like Steve was suggesting?

steveax:
@barriserloth, that can be a separate prepare for release commit.

steveax (approved):
👍

PR #234, opened 2020-10-13
(PCP-900) Throttle connection resets after crl is expired

justinstoller:
Previously, in 44ce091, we began closing all connections after a CRL
update to prevent stale SSLContexts from being held in memory. However,
the previous change dropped all connections in quick succession and
there was some concern that it could cause undo spikes in load from a
thundering herd of reconnection attempts.

This patch creates a background thread (the same as how we send inventory
updates) that every X milliseconds tries to close any Connections in the
inventory marked expired. It also throttles the attempt to close all
connections by sleeping Y milliseconds between each Connection it
closes. X and Y are configurable by the new config values
`crl-check-period` and `expired-conn-throttle` respectively. The default
to 60 seconds between CRL checks and 30 milliseconds between individual
Connection closes.

This should give a small but meaningful window for multiple CRL updates
to coalesce and cause connection resets for a 1000-connections-per-broker
deployment to take around 1 minute. Note, the `crl-check-period` is waited
for after/in-additon-to whatever time it takes to close any existing
expired connections.

Previously, the file-system watcher which used to trigger the connection
closing on CRL update now only marks existing Connections as expired for
the background thread to purge in its configured time.

### DIFF
--- a/src/puppetlabs/pcp/broker/core.clj
+++ b/src/puppetlabs/pcp/broker/core.clj
@@ -517,7 +517,7 @@
      :else
      ;; Generate an implicit association request and authorize association.
      (let [uri (ws->uri ws)
-           connection (connection/make-connection ws codec uri)
+           connection (connection/make-connection ws codec uri false)
            message (message/make-message
                      {:target "pcp:///server"
                       :sender uri
@@ -692,17 +692,26 @@
       :info {}
       (fn [_] (i18n/trs "Evicted all clients as there is no controller connection.")))))
 
-(defn restart-client-connections!
+(defn close-expired-connections!
   "Forces clients to reconnect because the broker's CRL is out of date"
   [broker]
-  (doseq [[_ client-connection] (:inventory @(:database broker))]
-    (websockets-client/close!
-      (:websocket client-connection)
-      1012 ;; code SERVER_RESTART, client *should* reconnect soon
-      (i18n/trs "CRL reloaded"))
-    (sl/maplog
-      :info {}
-      (fn [_] (i18n/trs "Evicted all clients because of CRL reload")))))
+  (let [inventory-snapshot (:inventory @(:database broker))
+        throttle-duration (:expired-conn-throttle broker)]
+    (sl/maplog :debug
+               {:count (count inventory-snapshot)}
+               #(i18n/trs "Checking the existing {0} connections for expired CRLs." (:count %)))
+    (doseq [[uri client-connection] inventory-snapshot]
+      (when (:expired client-connection)
+        (do
+          (sl/maplog :debug {:uri uri} #(i18n/trs "Closing expired connection for {0}." (:uri %)))
+          (Thread/sleep throttle-duration)
+          (websockets-client/close!
+            (:websocket client-connection)
+            1012 ;; code SERVER_RESTART, client *should* reconnect soon
+            (i18n/trs "CRL reloaded"))
+          (sl/maplog
+            :info {}
+            (fn [_] (i18n/trs "Evicted stale client connections because of CRL reload."))))))))
 
 (defn on-controller-connect!
   [broker controller-uri ws]
@@ -760,7 +769,7 @@
                ;; 0 : uri identifying connection
                ;; 1 : url to connect to
                #(i18n/trs "Connecting to {0} at {1}." (:pcpuri %) (:uri %)))
-    [pcp-uri (connection/make-connection client identity-codec pcp-uri)]))
+    [pcp-uri (connection/make-connection client identity-codec pcp-uri false)]))
 
 (s/defn initiate-controller-connections :- {p/Uri Connection}
   "Create PCP Clients for each controller URI"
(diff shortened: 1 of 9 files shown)


justinstoller:
This provides the spacing out of connection resets that @MikaelSmith encouraged in this [comment](https://github.com/puppetlabs/pcp-broker/pull/228#issuecomment-657734862).

justinstoller:
Gah, I forgot that schema validation only happens when running with a special profile, fixing.

justinstoller:
fwiw, I looked into one significantly different approach and one other implementation of this approach.

A different approach I researched was creating a custom onPong event handler to close an expired connection when it did it's ping-pong interaction. However Jetty _really_ doesn't want users writing custom onPong handlers and doing so looked like it might require reworking both pcp-broker and tk-jetty9.

A different implementation could have used tk-scheduler, however since we were already doing something very similar with a `future` for inventory updates, I went with this implementation.

justinstoller:
Fixed the schema testing, I'd love a quick look from someone on this.

justinstoller:
Also, if folks want to look at the logs the primary/console is "grey-expression" and I have six agents connected to the compiler "diehard-frog" (all in the pooler).

McdonaldSeanp (approved):

src/puppetlabs/pcp/broker/core.clj
```diff
@@ -692,17 +692,26 @@
...
-      1012 ;; code SERVER_RESTART, client *should* reconnect soon
-      (i18n/trs "CRL reloaded"))
-    (sl/maplog
-      :info {}
-      (fn [_] (i18n/trs "Evicted all clients because of CRL reload")))))
+  (let [inventory-snapshot (:inventory @(:database broker))
+        throttle-duration (:expired-conn-throttle broker)]
+    (sl/maplog :debug
+               {:count (count inventory-snapshot)}
+               #(i18n/trs "Checking the existing {0} connections for expired CRLs."))
```
MikaelSmith, line 702: I don't think this log invocation is right, it needs to pass an argument for `{0}` to `i18n/trs` as done at line 706.

src/puppetlabs/pcp/broker/service.clj
```diff
@@ -23,16 +23,20 @@
...
-                                             :authorization-check   authorization-check
-                                             :get-metrics-registry  get-metrics-registry
-                                             :max-connections       max-connections
-                                             :max-message-size      max-message-size
-                                             :idle-timeout          idle-timeout
-                                             :get-route             (partial get-route this)})]
+        (let [max-connections       (get-in-config [:pcp-broker :max-connections] 0)
+              max-message-size      (get-in-config [:pcp-broker :max-message-size] (* 64 1024 1024))
+              idle-timeout          (get-in-config [:pcp-broker :idle-timeout] (* 1000 60 6))
+              expired-conn-throttle (get-in-config [:pcp-broker :expired-conn-throttle] 30)
```
MikaelSmith, line 29: Is this 30ms? That doesn't seem to match the commit message.

MikaelSmith: We should update https://github.com/puppetlabs/pcp-broker/blob/master/doc/configuration.md with these new config options.

test/unit/puppetlabs/pcp/broker/core_test.clj
```diff
@@ -72,19 +72,28 @@
...
     (with-redefs [ws->uri mock-ws->uri]
       (let [broker (make-test-broker)
-            connection (connection/make-connection :dummy-ws identity-codec mock-uri)]
+            connection (connection/make-connection :dummy-ws identity-codec mock-uri false)]
         (swap! (:database broker) update :inventory assoc mock-uri connection)
         (is (not= {} (-> broker :database deref :inventory)))
         (remove-connection! broker mock-uri)
         (is (= {} (-> broker :database deref :inventory)))))))
 
+(deftest expire-connection-test
```
MikaelSmith, line 88: A test that expired connections are removed from the inventory would be handy.

MikaelSmith (changes requested):
Approach looks good. Pointed out a few things to fix.

src/puppetlabs/pcp/broker/service.clj
justinstoller: Sorry, I noticed that and went back and updated the commit message, but forgot to update the GH PR description.

justinstoller: Thanks for calling out the docs updates, missed those for the idle-timeout work as well.

test/unit/puppetlabs/pcp/broker/core_test.clj
justinstoller: I'm adding a test to show that they have `websockets/close!` called on them. Do you think that's sufficient, or should I validate that closing a connection does eventually trigger an inventory update?

MikaelSmith: That's enough for me.

justinstoller:
Thanks for the review, I think I've addressed the feedback.

test/unit/puppetlabs/pcp/broker/core_test.clj
```diff
@@ -72,19 +72,40 @@
...
+(deftest close-expired-connection-test
+  (testing "It should close expired connections in the inventory"
+    (let [closed (promise)]
+      (with-redefs [ws->uri mock-ws->uri
+                    puppetlabs.experimental.websockets.client/close! (fn [& args] (deliver closed true))]
+        (let [broker (make-test-broker)
+              connection (connection/make-connection :dummy-ws identity-codec mock-uri false)]
+          (add-connection! broker connection)
+          (expire-ssl-connections broker)
+          (close-expired-connections! broker)
```
MikaelSmith, line 106: Could this have a timing component? The close will happen up to 10ms after `expire-ssl-connections`.

justinstoller: Like increasing the throttle time and validating that it is honored?

MikaelSmith: I was thinking more that the test might fail, unless I'm forgetting how `realized?` works.

justinstoller: Or waiting on the promise with a timeout to ensure this doesn't get racey?

justinstoller: I should probably just do both  ( ;

justinstoller: Updated to check both sides of the wait, let me know if I missed something.

MikaelSmith (approved):

PR #236, opened 2020-10-15
(maint) Coerce URI to str before passing to sl/maplog

justinstoller:
The structured logging macro `maplog` will attempt to serialize a
URI, however we need its string representation for our logging. Coerce,
URIs to strings when closing connections due to CRL updates.


justinstoller:
I don't know why, but grokking & debugging `maplog` was one of hardest parts of this feature for me.

McdonaldSeanp (approved):

PR #237, opened 2020-10-19
(PE-30041) Add rpc_*_requests to default allowed message types

donoghuc:
When the controller-whitelist setting was moved to controller-allowlist it brought up a discussion around what reasonable defaults to expect for the setting. This commit updates the defaults to include the rpc messages where previously they were not included.


barriserloth:
Seems like there are some PR testing issues here

donoghuc:
I dont think my change broke those tests. They have failed before.

PR #239, opened 2020-12-11
(maint) Fix lein uberjar

Sharpie:
This commit adds a Bouncycastle dependency to the uberjar profile that
was missed when Bouncycastle usage was re-factored to support a
FIPS/non-FIPS split. The dependency allows `lein uberjar` to complete
instead of failing during AOT compilation.


trevor-vaughan:
FWIW, this works for me. :+1:

steveax (approved):
👍

PR #246, opened 2022-01-31
(PE-33424) Add Snyk scan

rileynewton:


.github/workflows/snyk_pr.yaml
```diff
@@ -0,0 +1,26 @@
+---
+name: snyk_pr
+
+on: pull_request_target
```
steveax, line 4: I think this leaks the snyk tokens to forks?

rileynewton: Hmm, Jeremy still has it in his [public workflow example](https://github.com/puppetlabs/security-snyk-clojure-action/blob/main/public_sample_workflow.yaml) - @LivingInSyn is this ok?

donoghuc: YEah `pull_request_target` is used explicitly for public repos. see conclusion of this thread https://puppet.slack.com/archives/CKMQ47015/p1639684402227100?thread_ts=1639683704.222900&cid=CKMQ47015

.github/workflows/snyk_pr.yaml
```diff
@@ -0,0 +1,28 @@
+---
+name: snyk_pr
+
+on:
+  pull_request_target
```
steveax, line 5: Needs a `:` on the end there (`pull_request_target:`)

.github/workflows/snyk_pr.yaml
rileynewton: Added conditional that makes this safe

steveax (approved):

.github/workflows/snyk_merge.yaml
```diff
@@ -0,0 +1,26 @@
+---
+name: snyk_merge
+
+on: push
```
LivingInSyn, line 4: be careful here, this run and push to the snyk console on _all_ pushes to all branches. Might not be a problem in your workflow but could be messy in others

.github/workflows/snyk_merge.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+ snyk_scan:
+   runs-on: ubuntu-latest
+   steps:
+    - name: checkout the current PR
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
LivingInSyn, line 13: Probably only want `fetch-depth: 1` here, original examples had 0 but that was wrong, we only need the latest version

.github/workflows/snyk_merge.yaml
```diff
@@ -0,0 +1,26 @@
...
+        persist-credentials: false
+    - name: Run Clojure Snyk Scan
+      id: scan
+      uses: puppetlabs/security-snyk-clojure-action@main
+      with:
+        snykToken: ${{ secrets.SNYK_PE_TOKEN }}
```
LivingInSyn, line 19: changed the token to `SNYK_FOSS_TOKEN`

.github/workflows/snyk_pr.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+ snyk_scan:
+   runs-on: ubuntu-latest
+   if: contains(github.event.pull_request.labels.*.name, 'safe to test')
+   steps:
+    - name: checkout the current PR
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
LivingInSyn, line 16: same thing as above, `fetch-depth: 1`

.github/workflows/snyk_merge.yaml
rileynewton: I think this is okay as generally snyk_merge will only run when a PR is completed because our workflow is generally to fork repos into our own private github account then do a PR of that branch, the github action that runs all the time is actually snky_pr but I think we want that. For a public repo like this that won't run all the time due to the labeling requirement.

rileynewton: @LivingInSyn I also think the 'safe to test' label credential should be added here as well?

.github/workflows/snyk_merge.yaml
rileynewton: Changed

.github/workflows/snyk_merge.yaml
rileynewton: @LivingInSyn I'm just going to drop synk_merge I think, I think @Magisus is right in doing that as what we want is already covered by snyk_pr.

.github/workflows/snyk_pr.yaml
```diff
@@ -0,0 +1,29 @@
+---
+name: snyk_pr
+
+on:
+  pull_request_target:
```
Magisus, line 5: Heads up, something I was worried about is true: This will run on the branch the PR is based on, and not on the contents of the PR. The alternative, using `pull_request`, means this job will fail if the PR is made from a fork. We should probably decide what is more important to us in this tradeoff.

Magisus: See https://github.com/puppetlabs/orchestrator/pull/1959 for an example of a change that should fix the security issue, but the job still fails.

rileynewton: Wow, very confusing stuff 😬

rileynewton: Leaving `pull-request-target` for now.

donoghuc (approved):

PR #249, opened 2023-07-07
(PE-36362) Update pcp-broker to use Jetty10

rileynewton:


test/integration/puppetlabs/pcp/broker/controllers_test.clj
```diff
@@ -304,53 +304,53 @@
...
+                         (is (= 1 (count (:warning-bin @database)))))
+                       (testing "clients connected until timeout is reached"
+                         (is (websocket-session/connected? (:websocket client-connection)))
+                         (testing "new connections are rejected while controllers in the bin"
+                           (with-open [client' (client/connect :certname agent2-cert)]
+                             (is (received? (client/recv! client')
+                                            [1011 "All controllers disconnected."]))))
+                         (deliver timed-out? true)
+                         (is (received? (client/recv! client)
+                                        [1001 "Shutdown"]))
```
rileynewton, line 335: I left this one assertion failing as I want to look into more why `1001 "Shutdown"` is no longer received, `1011 "All controllers disconnected."` is received instead.

test/integration/puppetlabs/pcp/broker/subscription_test.clj
```diff
@@ -143,17 +143,19 @@
                 (fn [broker client connection pattern-sets]
                   (let [result (subscribe-client! broker client connection pattern-sets)]
                     (deliver @subscription-finished true)
-                    (deref finish-subscription)
+                    ;; CODEREVIEW: This oddly breaks this test by the first controller not closing
+                    ;; properly, I don't understand why this behavior would have changed but
+                    ;; since it is a fairly artificial situation it's perhaps okay to remove?
+                    ;;(deref finish-subscription)
```
rileynewton, line 149: Could be something to look at more too, if this portends an actual issue.

rileynewton: More specifically this is what happens, the connection process to the new controller (client) begins before closing the first controller, leading pcp to error with that uri still being in inventory (note the "removing-connection" message at the end, that should be done before the second "HERE in connect!" and then there would be no node already associated error):

```
2023-07-05 17:02:17,998 DEBUG [p.t.internal] Lifecycle worker completed :boot lifecycle task; awaiting next task.
HERE in connect!

2023-07-05 17:02:18,464 DEBUG [o.e.j.w.c.i.WebSocketCoreSession] receiveFrame(TEXT@5c670111[len=154,fin=true,rsv=000,m=00000000]DirectByteBuffer@7fe88591[p=0,l=154,c=154,r=154]={<<<{"id":"bae023e4-999a-4de0...nt"],"subscribe":true}}>>>}, org.eclipse.jetty.websocket.core.internal.WebSocketConnection$1@65a2499b) - connectionState=WebSocketSessionState@12174644{OPEN,i=NO-OP,o=NO-OP,c=null}, handler=JettyWebSocketFrameHandler@3e5009dc[puppetlabs.trapperkeeper.services.webserver.jetty10_websockets.proxy$org.eclipse.jetty.websocket.api.WebSocketAdapter$CertGetter$2448cc42]
2023-07-05 17:02:18,481 INFO  [p.p.b.pcp_access] AUTHORIZATION_SUCCESS127.0.0.1:50266controller01.example.compcp://controller01.example.com/agenthttp://puppetlabs.com/inventory_requestbae023e4-999a-4de0-9b16-85c1207b6811pcp:///server
CLOSING CHANCLIENT!
FIRST CONTROLLER DONE

2023-07-05 17:02:22,975 DEBUG [o.q.c.QuartzSchedulerThread] batch acquisition of 1 triggers
HERE in connect!

2023-07-05 17:02:23,607 INFO  [p.p.b.core] Node with URI pcp://controller01.example.com/agent already associated with connection controller01.example.com 127.0.0.1:50266.
2023-07-05 17:02:23,617 INFO  [p.p.b.core] pcp://controller01.example.com/agent disconnected 1005 null
removing-connection:  pcp://controller01.example.com/agent
```

So presumably the code before _did_ remove the first controller from inventory before the second one started to connect. Probably worth confirming that. It's not exactly a race condition either in that sleeping doesn't help.

test/integration/puppetlabs/pcp/broker/controllers_test.clj
jonathannewman: If there are differences in behavior we really need to understand why that is the case.  These tests are set up with specific state expectations, and if we aren't reproducing all the states, that will likely be an issue.

rileynewton:
Added some CODEREVIEW comments to the test I wasn't able to figure out and a few other things that should be paid attention - lein test :only puppetlabs.pcp.broker.subscription-test/inventory-update-after-reconnect is the impossible test.

project.clj
```diff
@@ -13,16 +14,20 @@
   ;; requires lein 2.2.0+.
   :pedantic? :abort
 
-  :parent-project {:coords [puppetlabs/clj-parent "5.2.9"]
+  :parent-project {:coords [puppetlabs/clj-parent "7.0.1"]
```
jonathannewman, line 17:
```suggestion
  :parent-project {:coords [puppetlabs/clj-parent "7.2.7"]
```

project.clj
```diff
@@ -13,16 +14,20 @@
...
+
                    :inherit [:managed-dependencies]}
 
   :dependencies [[org.clojure/clojure]
                  [org.clojure/tools.logging]
+                 [hato "0.9.0"]
```
jonathannewman, line 23: If this isn't needed but for development, it should be in the dev dependencies.

project.clj
rileynewton: Should it be "8.0.0" actually once released?

project.clj
rileynewton: Moved

test/integration/puppetlabs/pcp/broker/subscription_test.clj
rileynewton: This test failure still exists 😢

project.clj
jonathannewman: Not sure we have to do a major version number release as we are adding a new dependency, not removing any.

project.clj
```diff
@@ -13,16 +14,19 @@
...
                  [org.clojure/tools.logging]
                  [puppetlabs/kitchensink]
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/trapperkeeper-authorization]
-                 [puppetlabs/trapperkeeper-metrics]
-                 [puppetlabs/trapperkeeper-webserver-jetty9]
+                 [puppetlabs/trapperkeeper-metrics "2.0.0"]
+                 [com.puppetlabs/trapperkeeper-webserver-jetty10 "1.0.2"]
+                 [org.eclipse.jetty.websocket/websocket-jetty-client ~jetty-version]
+                 [org.eclipse.jetty.websocket/websocket-jetty-api ~jetty-version]
```
jonathannewman, line 29: Are these dev dependencies?

rileynewton:
There does appear to be an intermittent test failure with the following returning a nil associate response when run as part of the whole test suite:

```
lein test :only puppetlabs.pcp.broker.service-test/client-type-must-match-for-association-test

FAIL in (client-type-must-match-for-association-test) (service_test.clj:78)
Unsuccessful associate_response and WebSocket closes if client requests new association with version: "v2.0"
expected: (= "http://puppetlabs.com/associate_response" (:message_type message))
  actual: (not (= "http://puppetlabs.com/associate_response" nil))

```

So the remaining failing tests are that one intermittently and `lein test :only puppetlabs.pcp.broker.subscription-test/inventory-update-after-reconnect`

project.clj
rileynewton: Good point.. I think they can be removed alltogether since clj-pcp-client includes them.

rileynewton: websocket-jetty-api is used in non-test code here https://github.com/puppetlabs/pcp-broker/blob/9e644238664ede59e4628bb8493fafc464d9047c/src/puppetlabs/pcp/broker/websocket.clj#L7 - but client is only used for testing.

project.clj
rileynewton: We technically are removing tk-jetty-9, replacing with tk-jetty-10, but yeah not sure it really needs a major version bump.

rileynewton: Updated to 7.2.7

project.clj
rileynewton: Removed them

test/integration/puppetlabs/pcp/broker/controllers_test.clj
rileynewton: This different code is resolved now with the new client, it receives 1001 as expected the message has just changed to "1001 "Connection Idle Timeout"" -- underneath it is still Shutdown.

project.clj
jonathannewman: I think we will keep jetty9 in there until everything is migrated over, so there will be a period of time where both are used.

jonathannewman:
the bouncy castle references in project.clj need to be updated to `18on` from `15on`

test/integration/puppetlabs/pcp/broker/service_test.clj
```diff
@@ -250,20 +248,23 @@
...
                                   {:message_type "http://puppetlabs.com/inventory_request"
                                    :target "pcp:///server"
                                    :sender "pcp://client01.example.com/test"
                                    :data {:query ["pcp://client01.example.com/test"]}})]
-                     (client/send! client request)
-                     (let [response (client/recv! client 1000)]
+                     (.send client request)
+                     ;; CODEREVIEW: sendString in pcp-client throws an IOException here, apparently 
+                     ;; this didn't used to happen? We're using the blocking send so 
+                     ;; perhaps this needs to be addressed in pcp-client.
```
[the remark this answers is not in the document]
rileynewton: The pcp-client production code has a try/catch though so hopefully we're alright: https://github.com/puppetlabs/clj-pcp-client/blob/1116cd18ee72cfde9096fe416f034cce139a8fbd/src/puppetlabs/pcp/client.clj#L236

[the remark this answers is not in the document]
rileynewton: Hmm it catches ExecutionException though.. might not be right

jonathannewman (approved):

PR #254, opened 2023-11-27
(maint) add ci-profile, remove duplicate dependency

jonathannewman:
This adds a ci profile, and removes the duplicate dependency for pcp-common which comes from the pcp-client.


rileynewton (approved):
👍

PR #256, opened 2023-12-07
(maint) handle connection exceptions

jonathannewman:
This adds some exception handling to the connection setup to allow exceptions to be logged.  Prior to this change, an exception would stop the broker from correctly starting.

This also adds some additional logging to help diagnose issues.


rileynewton (approved):
👍

PR #257, opened 2024-02-12
(maint) update clj-parent (7.3.7), remove pins

steveax:
This commit updates clj-parent to 7.3.6 and removes the pins for tk-jetty10 and tk-metrics.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -13,7 +13,7 @@
   ;; requires lein 2.2.0+.
   :pedantic? :abort
 
-  :parent-project {:coords [puppetlabs/clj-parent "7.2.7"]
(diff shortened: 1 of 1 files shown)


rileynewton (approved):
👍

jonathannewman:
I think https://github.com/puppetlabs/clj-parent/pull/451 should go in first, and this update to 7.3.7  Unfortunately version 2.0.1 of metrics pins the tk-ws-jetty10 version, so we need 2.0.2, which is brought in by that.

jonathannewman:
Ok, clj-parent 7.3.7 is out now.

jonathannewman (approved):

